PR #9, opened 2025-10-02
Error type for proper error handling and bucket name validation

### LINKED ISSUE #6: Error handling
Similar to how we have specific errors and errors codes in RHP4, S3 also got its own codes and identifiers for errors. 
These errors are returned as an xml response. 

We should draw some inspiration from [here](https://github.com/SiaFoundation/gofakes3/blob/master/error.go). and implement something similar. Depending on the implementation we might also want to copy it since 90% of it is just defining the error codes and types that S3 expects.

ChrisSchinnerl:
This PR adds the `S3Error` type and makes use of it in the existing code.
It also extends the bucket test to make sure that the bucket validation works which is also added in this PR.

Fixes https://github.com/SiaFoundation/s3d/issues/6

### DIFF
--- a/s3/buckets.go
+++ b/s3/buckets.go
@@ -30,7 +30,7 @@ func (s *s3) routeBucket(w http.ResponseWriter, r *http.Request, bucket string) 
 			return errors.New("createObjectBrowserUpload is not implemented")
 		}
 	default:
-		panic("TODO: implement S3 compliant error handling")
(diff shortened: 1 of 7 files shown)


peterjan (approved):

peterjan:
@ChrisSchinnerl S3 compat tests are failing it seems

PR #10, opened 2025-10-03
Basic signature authentication

ChrisSchinnerl:
This PR adds basic v4 authentication. "Basic" in the sense that it works for non-chunked requests.
It is a lot of code but large chunks of it are constants and descriptions in `auth.go` or modified parsing and canonical request creation code from our gofakes3 fork in `auth_v4.go` and `parse_v4.go`.  

Since authentication requires quite a bit of new code I moved it into its own package which also lead to errors getting their own package to avoid cycles between `s3` and `auth`.

Successful authentication is tested against the official AWS Go SDK for both `CreateBucket` and `ListBucket` right now since those are the only supported routes at the moment. Future endpoints tested with the `S3Tester` helper type will default to authenticating using test credentials.


s3/auth.go
```diff
@@ -0,0 +1,89 @@
...
+	ContentStreamingAWS4HMACSHA256PayloadTrailer = "STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER"
+
+	ContentStreamingAWS4ECDSAP256SHA256Payload = "STREAMING-AWS4-ECDSA-P256-SHA256-PAYLOAD"
+
+	ContentStreamingAWS4ECDSAP256SHA256PayloadTrailer = "STREAMING-AWS4-ECDSA-P256-SHA256-PAYLOAD-TRAILER"
+)
+
+// authenticatedHandler is like http.Handler but includes the access key ID of
+// the authenticated user.
+type authenticatedHandler interface {
```
ChrisSchinnerl: This might be a bit more verbose than attaching user info to the context but I do appreciate being explicit here.

s3/auth/auth.go
```diff
@@ -0,0 +1,110 @@
...
+// authenticatedHandlerFunc(f) is an authenticated handler that calls f.
+type AuthenticatedHandlerFunc func(http.ResponseWriter, *http.Request, *string)
+
+// ServeHTTP calls f(w, r, accessKeyID).
+func (f AuthenticatedHandlerFunc) ServeHTTP(w http.ResponseWriter, r *http.Request, accessKeyID *string) {
+	f(w, r, accessKeyID)
+}
+
+// HandleAuth inspects the request to determine the authentication type, verfies
+// the signature returns the used access key ID.
```
peterjan, line 74: typo
```suggestion
// the signature and returns the used access key ID.
```

s3/auth/auth.go
```diff
@@ -0,0 +1,110 @@
...
+//
+// - 'now' refers to the current time and is used to verify request timestamps
+// - 'region' is the AWS region the request is targeted to. If the region is the
```
peterjan, line 77:
```suggestion
// - 'region' is the AWS region the request is targeted to. If the region is an
```

s3/s3.go
```diff
@@ -181,44 +223,49 @@ func (s *s3) routeBase(w http.ResponseWriter, r *http.Request) {
...
-	http.Error(w, "routeObject is not implemented", http.StatusNotImplemented)
-	return nil
+	return s3errs.ErrNotImplemented
+}
+
+// assertAuth checks if the accessKeyID is not nil, returning an error if it is.
+// If the accessKeyID is valid, it is returned as a string. This adds a layer of
+// safety to ensure that handlers that require authentication are not
+// accidentally called with an empty accessKeyID.
+func assertAuth(accessKeyID *string) (string, error) {
```
peterjan, line 264: I'm a little confused why we have handlers that take a pointer, which then all immediately call `assertAuth` that returns an s3 error if it's `nil`. We have control over the routing right? So why don't have handlers that require authentication take an accessKeyID, and return `s3errs.ErrAccessDenied` in the routing?

peterjan (approved):

s3/s3.go
ChrisSchinnerl: Mostly because routing in S3 is a pain. It's all just a single route with various combinations of query params. For some routes it is even optional.
The big difference between `gofakes3` and this is that we push more of the responsibility when it comes to access controls on the backend. So the backend receives a pointer to an access key id and gets to decide itself whether that is alright or not.

PR #13, opened 2025-10-21
GetObject and HeadObject

ChrisSchinnerl:
This PR implements `GetObject` and `HeadObject`

### DIFF
--- a/s3/internal/testutil/testutil.go
+++ b/s3/internal/testutil/testutil.go
@@ -221,7 +221,7 @@ func rangeHeader(rnge *s3.ObjectRangeRequest) *string {
 	}
 	var s string
 	if rnge.FromEnd {
-		s = fmt.Sprintf("bytes=-%d", rnge.End)
+		s = fmt.Sprintf("bytes=-%d", rnge.Start)
 	} else if rnge.End == -1 {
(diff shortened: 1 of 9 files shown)


s3/internal/testutil/testutil.go
```diff
@@ -35,6 +47,73 @@ func (t *S3Tester) CreateBucket(ctx context.Context, bucket string) error {
...
+func (t *S3Tester) GetObject(ctx context.Context, bucket, object string, rnge *s3.ObjectRangeRequest) (*s3.Object, error) {
+	resp, err := t.client.GetObject(ctx, &service.GetObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(object),
+		Range:  rangeHeader(rnge),
+	})
+	if err != nil {
+		return nil, err
+	}
+	etag := strings.Trim(*resp.ETag, `"`)
```
peterjan, line 60: can we safely deref `resp.ETag` here?

peterjan (approved):
LGTM 👍

s3/internal/testutil/testutil.go
ChrisSchinnerl: Should be safe since we always set it on the backend.

PR #14, opened 2025-10-23
PutObject

ChrisSchinnerl:
This PR adds support for basic `PutObject` without versioning

### DIFF
--- a/s3/s3.go
+++ b/s3/s3.go
@@ -2,6 +2,7 @@ package s3
 
 import (
 	"context"
+	"io"
 	"net"
 	"net/http"
 	"strings"
@@ -48,6 +49,20 @@ type Backend interface {
 	// ListBuckets lists all available buckets for the user identified by the
 	// given access key.
 	ListBuckets(ctx context.Context, accessKeyID string) ([]BucketInfo, error)
+
+	// PutObject puts an object with the given key into the specified bucket.
+	//
+	// - If the access key does not have permission to store the object,
+	//   [ErrAccessDenied] must be returned.
+	//
+	// - If the bucket does not exist, [ErrNoSuchBucket] must be returned.
+	//
+	// - If the object already exists, it must be overwritten.
+	//   NOTE: Versioning is not supported yet.
+	//
+	// - If the bytes read from 'r' do not match 'contentLength',
+	// [ErrIncompleteBody] must be returned.
+	PutObject(ctx context.Context, accessKeyID string, bucket, object string, meta map[string]string, r io.Reader, contentLength int64) ([]byte, error)
 }
 
 type s3 struct {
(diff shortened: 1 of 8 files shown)


s3/s3.go
```diff
@@ -48,6 +49,20 @@ type Backend interface {
 	// ListBuckets lists all available buckets for the user identified by the
 	// given access key.
 	ListBuckets(ctx context.Context, accessKeyID string) ([]BucketInfo, error)
+
+	// PutObject puts an object with the given key into the specified bucket.
```
peterjan, line 53: no mention of `ErrMetadataTooLarge`?

s3/s3.go
```diff
@@ -48,6 +49,20 @@ type Backend interface {
...
+	//
+	// - If the access key does not have permission to store the object,
```
peterjan, line 55: no mention of `ErrKeyTooLongError`?

s3/s3.go
```diff
@@ -48,6 +49,20 @@ type Backend interface {
...
+	//   [ErrAccessDenied] must be returned.
+	//
+	// - If the bucket does not exist, [ErrNoSuchBucket] must be returned.
+	//
+	// - If the object already exists, it must be overwritten.
+	//   NOTE: Versioning is not supported yet.
+	//
+	// - If the bytes read from 'r' do not match 'contentLength',
+	// [ErrIncompleteBody] must be returned.
+	PutObject(ctx context.Context, accessKeyID string, bucket, object string, meta map[string]string, r io.Reader, contentLength int64) ([]byte, error)
```
peterjan, line 65: Is there a limit her eon object size? (Very) quick google says AWS imposes a 5GB limit.

peterjan (approved):

s3/s3.go
ChrisSchinnerl: The backend doesn't need to care about it. It's thrown before `PutObject` is called on the `backend` interface.

s3/s3.go
ChrisSchinnerl: Same as the other error. It's caught before `PutObject` is called. It doesn't work for all errors which is why there are some listed on the interface but generally any error that can be caught before calling out to the backend is one error the implementation of the interface doesn't need to care about

s3/s3.go
ChrisSchinnerl: I haven't seen one on the `PutObject` documentation. Googling seems to say the limit for an object is 5TB and the recommendation for `PutObject` is 5GB. But I honestly don't know. There is also no dedicated error error for it from what I can tell. Do you have a link?

peterjan: https://docs.aws.amazon.com/AmazonS3/latest/userguide/upload-objects.html

> Upload an object in a single operation by using the AWS SDKs, REST API, or AWS CLI – With a single PUT operation, you can upload a single object up to 5 GB in size.

Looking at the [error list](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList). I think it should return a 400 Bad Request with error code `EntityTooLarge`.

peterjan: Fine with not doing this though but it does seem a little sketchy not having a limit imposed there.

PR #22, opened 2025-10-27
ListObjects

ChrisSchinnerl:
This implements the V2 version of ListObjects and `ListObjectsVersion`. The latter is just a wrapper for `ListObjects` since we don't support versioning. 

The PR is unfortunately quite long but a lot of it is test utils code for the prefix matching logic and xml types for the response messages.

### DIFF
--- a/s3/messages.go
+++ b/s3/messages.go
@@ -104,6 +104,7 @@ type (
 		StartAfter string `xml:"StartAfter,omitempty"`
 	}
 
+	// Version represents a version of an S3 object.
 	Version struct {
 		XMLName      xml.Name    `xml:"Version"`
 		Key          string      `xml:"Key"`
@@ -119,6 +120,7 @@ type (
 		Owner *UserInfo `xml:"Owner,omitempty"`
 	}
 
+	// ListObjectVersionsResult is the response to a ListObjectVersions request.
 	ListObjectVersionsResult struct {
 		XMLName        xml.Name       `xml:"ListObjectVersionsResult"`
 		Xmlns          string         `xml:"xmlns,attr"`
--- a/s3/messages.go
+++ b/s3/messages.go
@@ -73,6 +73,7 @@ type (
 		Prefix string `xml:"Prefix"`
 	}
 
+	// Content represents an S3 object in a bucket listing.
 	Content struct {
 		Key          string       `xml:"Key"`
 		LastModified ContentTime  `xml:"LastModified"`
@@ -82,6 +83,7 @@ type (
 		Owner        *UserInfo    `xml:"Owner,omitempty"`
 	}
 
+	// ListObjectsV2Result is the response to a ListObjectsV2 request.
 	ListObjectsV2Result struct {
 		ListObjectsResultBase
 
@@ -102,6 +104,7 @@ type (
 		StartAfter string `xml:"StartAfter,omitempty"`
 	}
 
+	// ListObjectsResultBase is the common part of a listing response.
 	ListObjectsResultBase struct {
 		XMLName xml.Name `xml:"ListBucketResult"`
 		Xmlns   string   `xml:"xmlns,attr"`
@@ -132,9 +135,13 @@ type (
 		Contents       []*Content     `xml:"Contents"`
 	}
 
+	// StorageClass represents the storage class of an S3 object. If not specified,
+	// it defaults to "STANDARD".
 	StorageClass string
 )
 
+// MarshalXML implements custom XML marshalling for StorageClass to override the
+// empty value.
 func (s StorageClass) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
 	if s == "" {
 		s = "STANDARD"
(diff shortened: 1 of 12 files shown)


s3/messages.go
```diff
@@ -65,3 +65,144 @@ type GetBucketLocation struct {
...
+		VersionIDMarker string `xml:"VersionIdMarker,omitempty"`
+
+		// When the number of responses exceeds the value of MaxKeys,
+		// NextVersionIdMarker specifies the first object version not returned that
+		// satisfies the search criteria. Use this value for the version-id-marker
+		// request parameter in a subsequent request.
+		NextVersionIDMarker string `xml:"NextVersionIdMarker,omitempty"`
+
+		// AWS responds with a list of either <Version> or <DeleteMarker> objects. The order
+		// needs to be preserved and they need to be direct of ListBucketVersionsResult:
```
peterjan: "... need to be direct of ..." typo ?

s3/messages.go
```diff
@@ -65,3 +65,144 @@ type GetBucketLocation struct {
...
+		// all of the results might not be returned.
+		IsTruncated bool `xml:"IsTruncated"`
+
+		// Causes keys that contain the same string between the prefix and the
+		// first occurrence of the delimiter to be rolled up into a single result
+		// element in the CommonPrefixes collection. These rolled-up keys are not
+		// returned elsewhere in the response.
+		//
+		// NOTE: Each rolled-up result in CommonPrefixes counts as only one return
+		// against the MaxKeys value. (BW: been waiting to find some confirmation of
```
peterjan: copy pasta of initials?

peterjan: fwiw it's correct, keys rolled up into a CommonPrefixes entry count as one toward MaxKeys

> [Delimiter](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html?utm_source=chatgpt.com#API_ListObjects_ResponseSyntax)
Causes keys that contain the same string between the prefix and the first occurrence of the delimiter to be rolled up into a single result element in the CommonPrefixes collection. These rolled-up keys are not returned elsewhere in the response. Each rolled-up result counts as only one return against the MaxKeys value.

s3/objects.go
```diff
@@ -260,6 +436,66 @@ func (o *ObjectRangeRequest) Range(size int64) (*ObjectRange, error) {
...
+
+	page.MaxKeys = maxKeys
+
+	if _, hasMarker := query["continuation-token"]; hasMarker {
+		// list Objects V2 uses continuation-token preferentially, or
+		// start-after if continuation-token is missing. continuation-token is
+		// an opaque value that looks like this: 1ueGcxLPRx1Tr/XYExHnhbYLgveDs2J/wm36Hy4vbOwM=.
+		// This just looks like base64 junk so we just cheat and base64 encode
+		// the next marker and hide it in a continuation-token.
+		tok, err := base64.URLEncoding.DecodeString(query.Get("continuation-token"))
```
peterjan, line 453: Do we have to handle both types of continuation tokens? Because this will properly decode a token we encoded ourselves, as we use `base64.URLEncoding`,  but we wouldn't be able to decode the token from the comment. (It uses `StdEncoding`, note the `/`)

s3/objects.go
```diff
@@ -151,6 +177,137 @@ func (s *s3) headObject(w http.ResponseWriter, r *http.Request, accessKeyID *str
...
+		},
+		KeyCount:          int64(len(objects.CommonPrefixes) + len(objects.Contents)),
+		StartAfter:        q.Get("start-after"),
+		ContinuationToken: q.Get("continuation-token"),
+	}
+	if objects.NextMarker != "" {
+		// S3 continuation tokens are opaque base64-like values, so we just
+		// base64 encode the next marker and hand it out as a continuation
+		// token.
+		result.NextContinuationToken = base64.URLEncoding.EncodeToString([]byte(objects.NextMarker))
```
peterjan, line 254: Should we use `StdEncoding`, see https://github.com/SiaFoundation/s3d/pull/22/files#r2486422728

peterjan (changes requested):
Not sure about the encoding of the continuation token, otherwise LGTM

s3/messages.go
ChrisSchinnerl: Not sure tbh. I kept the original comment on the gofakes3 type but this doesn't make a ton of sense to me.

s3/messages.go
ChrisSchinnerl: Indeed

s3/objects.go
ChrisSchinnerl: I'm not sure why but they are both listed for `ListObjectsV2` and it is easy enough to support both. Gofakes3 uses `URLEncoding` but I changed it to `StdEncoding`. Probably doesn't matter but might as well be able to parse the example.

ChrisSchinnerl: Ok nevermind. The token is std encoded but then you need to query escape it which is annoying. Might as well use URLEncoding since it's just base64 - like anyway.

peterjan: It's kind of sad we wouldn't be able to parse the continuation token example from the official documentation? Why don't we fallback and do 

```go
for _, decoder := range []*base64.Encoding{
	base64.URLEncoding,
	base64.StdEncoding,
} { ... }
```

ChrisSchinnerl: Because there is no point in falling back to an encoding we know won't result in a valid token anyway.

peterjan (approved):

PR #23, opened 2025-10-29
DeleteObjects

ChrisSchinnerl:
Adds `DeleteObjects` for deleting objects in bulk

### DIFF
--- a/s3/objects.go
+++ b/s3/objects.go
@@ -125,10 +125,10 @@ func (s *s3) deleteObjects(w http.ResponseWriter, r *http.Request, accessKeyID s
 		return s3errs.ErrMalformedXML
 	}
 
-	for _, obj := range req.Objects {
-		if obj.VersionID == "null" {
-			obj.VersionID = ""
-		} else if obj.VersionID != "" {
+	for i := range req.Objects {
+		if req.Objects[i].VersionID == "null" {
+			req.Objects[i].VersionID = ""
+		} else if req.Objects[i].VersionID != "" {
 			return s3errs.ErrNotImplemented // versioning not supported
 		}
 	}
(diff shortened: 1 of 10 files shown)


s3/objects.go
```diff
@@ -111,6 +112,38 @@ func (s *s3) deleteObject(w http.ResponseWriter, r *http.Request, accessKeyID st
...
+	if err := decodeXMLBody(r.Body, &req); err != nil {
+		return err
+	}
+
+	if len(req.Objects) > MaxBucketKeys {
+		return s3errs.ErrMalformedXML
+	}
+
+	for i := range req.Objects {
+		if req.Objects[i].VersionID == "null" {
```
peterjan: I've seen the `"null"` check a couple of times now, probably better to extract this sentinel value in a constant

s3/s3.go
```diff
@@ -47,6 +47,19 @@ type Backend interface {
...
+	// user identified by the given access key.
+	//
+	// - If the access key does not have permission to delete the objects,
+	//   [ErrAccessDenied] must be returned.
+	//
+	// - If the bucket does not exist, [ErrNoSuchBucket] must be returned.
+	//
+	// - If any of the objects with the given keys in the specified bucket do not
+	//   exist, they must be reported in the returned ObjectsDeleteResult using
+	//   the [ErrNoSuchKey] error.
```
peterjan: Hm looking at [the docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html) it says that deletions should be idempotent:

> For each key, Amazon S3 performs a delete operation and returns the result of that delete, success or failure, in the response. If the object specified in the request isn't found, Amazon S3 confirms the deletion by returning the result as deleted.

peterjan (changes requested):

peterjan (dismissed):

PR #26, opened 2025-10-30
Add support for `STREAMING-UNSIGNED-PAYLOAD-TRAILER` auth

ChrisSchinnerl:
Which seems to be the default for the AWS v2 Go SDK when https is used

### DIFF
--- a/s3/auth/streaming_v4.go
+++ b/s3/auth/streaming_v4.go
@@ -43,7 +43,7 @@ func handleAuthV4Streaming(req *http.Request) error {
 
 	switch req.Header.Get(HeaderXAMZContentSHA256) {
 	case ContentStreamingUnsignedPayloadTrailer:
-		req.Body = io.NopCloser(newChunkedPayloadTrailerReader(req.Body, expectedHeaders))
+		req.Body = newChunkedPayloadTrailerReader(req.Body, expectedHeaders)
 	case ContentStreamingAWS4HMACSHA256Payload:
 		return s3errs.ErrNotImplemented
 	case ContentStreamingAWS4HMACSHA256PayloadTrailer:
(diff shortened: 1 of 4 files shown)


s3/auth/streaming_v4.go
```diff
@@ -0,0 +1,287 @@
...
+
+	// parse the headers
+	expectedHeaders := make(map[string]struct{})
+	for h := range strings.SplitSeq(xAmzTrailer, ",") {
+		expectedHeaders[http.CanonicalHeaderKey(h)] = struct{}{}
+	}
+
+	switch req.Header.Get(HeaderXAMZContentSHA256) {
+	case ContentStreamingUnsignedPayloadTrailer:
+		req.Body = io.NopCloser(newChunkedPayloadTrailerReader(req.Body, expectedHeaders))
```
peterjan: 😬 can we do that (the `io.NopCloser`)? Especially for TLS it would be quite wasteful to not properly handle the underlying connection and hand it back to the pool? I'm not 100% on the inner workings but it's not that much effort to forward `Close` to the original body to ensure we don't leak connections

s3/auth/streaming_v4.go
```diff
@@ -0,0 +1,287 @@
...
+
+	// parse x-amz-trailer which contains the expected trailer headers
+	xAmzTrailer := req.Header.Get(HeaderXAMZTrailer)
+	if xAmzTrailer == "" {
+		return s3errs.ErrInvalidArgument
+	}
+
+	// parse the headers
+	expectedHeaders := make(map[string]struct{})
+	for h := range strings.SplitSeq(xAmzTrailer, ",") {
```
peterjan, line 40: Fancy

peterjan (changes requested):

peterjan (approved):

PR #27, opened 2025-10-31
Validate `x-amz-content-sha256` checksum if present

ChrisSchinnerl:
So far we have validated the signature on the signed request but not the payload in case the `x-amz-content-sha256` header contains the actual payload.

### DIFF
--- a/.github/workflows/tox-test.yml
+++ b/.github/workflows/tox-test.yml
@@ -43,12 +43,16 @@ jobs:
         run: |
           nohup go run ./cmd/s3d-tox/main.go > s3d.log 2>&1 &
           echo $! > s3d.pid
-          sleep 1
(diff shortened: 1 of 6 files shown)


.github/workflows/tox-test.yml
```diff
@@ -43,12 +43,16 @@ jobs:
...
-            cat s3d.log
-            exit 1
-          fi
+
+          # wait a bit to let s3d start
+          for i in {1..10}; do
+            if kill -0 $(cat s3d.pid); then
+              sleep 1
+              if nc -z localhost 8000; then
+                break
```
peterjan, line 52: we could set a variable here and check it after looping to print a meaningful error message in case `nc -z localhost 8000` is never true

peterjan (dismissed):

PR #29, opened 2025-11-03
Copy Object

ChrisSchinnerl:


s3/messages.go
```diff
@@ -248,3 +248,12 @@ type (
 		RequestID string   `xml:"RequestId,omitempty"`
 	}
 )
+
+type (
+	// ObjectCopyResult contains the response from a CopyObject operation.
+	ObjectCopyResult struct {
+		XMLName      xml.Name    `xml:"CopyObjectResult"`
+		ETag         string      `xml:"ETag,omitempty"`
```
peterjan: Is the `ETag` every empty? I think I noticed that in all PRs I've seen so far, that the `ETag` never has an `omitempty` and now it does.

s3/objects.go
```diff
@@ -113,6 +121,53 @@ func (s *s3) routeObject(w http.ResponseWriter, r *http.Request, accessKeyID *st
...
+		zap.String("source", source),
+	)
+	log.Debug("copy object")
+
+	if len(source) > KeySizeLimit {
+		return s3errs.ErrKeyTooLongError
+	}
+
+	// parse source
+	parts := strings.SplitN(strings.TrimPrefix(source, "/"), "/", 2)
```
peterjan, line 137: Do we have sufficient testing around `X-Amz-Copy-Source`?

s3/objects_test.go
```diff
@@ -212,6 +212,70 @@ func TestPutObject(t *testing.T) {
 	})
 }
 
+func TestCopyObject(t *testing.T) {
```
peterjan, line 215: could be extended to cover `ErrNoSuchKey`

s3/objects_test.go
```diff
@@ -212,6 +212,70 @@ func TestPutObject(t *testing.T) {
...
+	} else if !bytes.Equal(data, fetched) {
+		t.Fatal("data mismatch")
+	} else if len(obj.Metadata) != 2 || obj.Metadata["foo"] != "bar" || obj.Metadata["baz"] != "qux" {
+		t.Fatalf("metadata mismatch: %+v", obj.Metadata)
+	} else if obj.ContentMD5 != hash {
+		t.Fatal("hash mismatch", obj.ContentMD5, hash[:])
+	} else if obj.LastModified.Before(copyTime) {
+		t.Fatalf("last modified mismatch: expected after %v, got %v", copyTime, obj.LastModified)
+	}
+}
```
peterjan, line 304: add a case to cover 

>  If the source and destination are the same, the object is kept but its metadata
	//   is replaced with the provided metadata.

peterjan (changes requested):

s3/objects.go
ChrisSchinnerl: You are the reviewer. You tell me 😅

peterjan: I think it's fine because strings.SplitN returns the whole input string if it can't match, this [0] on line 142 is what caught my eye.

peterjan (dismissed):

PR #32, opened 2025-11-10
Enable and fix compatibility tests for v2 object listing

ChrisSchinnerl:
This PR enables the subset of `ListObjectsV2` tests in `test_s3.py` of the compatibility test suite and fixes issues that caused tests to fail.

### DIFF
--- a/.github/workflows/tox-test.yml
+++ b/.github/workflows/tox-test.yml
@@ -80,7 +80,8 @@ jobs:
           S3_USE_SIGV4: "True" # We only support v4 signatures
         run: |
           tox run --skip-pkg-install -c ./s3-tests -- s3tests/functional/test_s3.py \
-            -k "test_bucket_listv2 and not delimiter_percentage and not delimiter_whitespace"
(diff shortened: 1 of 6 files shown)


.github/workflows/tox-test.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           python -m pip install "tox>=4"
 
       - name: Clone test repo
-        run: git clone https://github.com/ceph/s3-tests.git
+        run: git clone https://github.com/SiaFoundation/s3-tests.git
```
ChrisSchinnerl, line 31: NOTE: I forked the repo to be able to add the `s3d_not_implemented` and `s3d_not_supported` markers. 

Since there are a ton of tests I have only applied them to the once relevant in this PR but the gist is that we can use those to exclude tests for features we either want to support and haven't implemented yet or features we don't want to support at all. Such as custom delimiters that are not `/`.

peterjan (approved):
Nice

PR #43, opened 2025-11-14
Add ListParts

### LINKED ISSUE #19: Implement multipart uploads
This includes

- [x] Creating the upload
- [x] Adding a part to it
- [x] Completing it
- [x] Cancelling it
- [x] Listing ongoing uploads
- [ ] List parts

peterjan:
Closes #19


s3/messages.go
```diff
@@ -271,4 +271,30 @@ type (
...
+	// ListPartsResponse matches the XML response returned by AWS when listing
+	// uploaded parts for an in-progress multipart upload.
+	ListPartsResponse struct {
+		XMLName              xml.Name             `xml:"ListPartsResult"`
+		Xmlns                string               `xml:"xmlns,attr"`
+		Bucket               string               `xml:"Bucket"`
+		Key                  string               `xml:"Key"`
+		UploadID             string               `xml:"UploadId"`
+		PartNumberMarker     int                  `xml:"PartNumberMarker"`
+		NextPartNumberMarker int                  `xml:"NextPartNumberMarker,omitempty"`
```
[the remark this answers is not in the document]
peterjan: Not sure what you are on about, the XML response in the docs define them as integers.

s3/multipart.go
```diff
@@ -166,3 +264,33 @@ func parsePartNumber(raw string) (int, error) {
...
+
+	if rawMarker := query.Get("part-number-marker"); rawMarker != "" {
+		val, err := strconv.Atoi(rawMarker)
+		if err != nil {
+			return page, s3errs.ErrInvalidArgument
+		}
+		if val < 0 || val >= MaxUploadPartNumber {
+			return page, s3errs.ErrInvalidArgument
+		}
+		page.PartNumberMarker = rawMarker
```
peterjan: This is a little odd, the [AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListParts.html) require an integer in the response but the markers are generally strings and `service.ListPartsOutput` also defines it using an aws string.

s3/multipart.go
```diff
@@ -42,6 +44,34 @@ type UploadPartResult struct {
 	ContentMD5 [16]byte
 }
 
+// ListPartsPage specifies pagination options when listing the parts of an
+// in-progress multipart upload.
+type ListPartsPage struct {
+	PartNumberMarker string
```
ChrisSchinnerl: Why not change this to `int`? It's not an exposed response type so we can change it and then get rid of all the redundant `Atoi` calls

s3/multipart.go
ChrisSchinnerl: The AWS SKD is awful and probably generated. I'd keep a proper integer around for as long as possible and only when passing it to the SDK turn it into a string.

ChrisSchinnerl (changes requested):

s3/multipart.go
peterjan: Stupid codex made me second guess and change it.

ChrisSchinnerl (dismissed):
Lgtm but got conflicts

ChrisSchinnerl (approved):

PR #45, opened 2025-11-17
Add buckets and objects

### LINKED ISSUE #44: Sia Backend: Implement bucket creation, deletion, PutObject and GetObject
Implement the requirements for basic uploads and downloads

ChrisSchinnerl:
Depends on https://github.com/SiaFoundation/indexd/pull/671

Closes https://github.com/SiaFoundation/s3d/issues/44


sia/persist/sqlite/buckets.go
```diff
@@ -0,0 +1,89 @@
...
+	})
+	return buckets, err
+}
+
+// bucketID returns the ID of the bucket with the given name or an error if it
+// does not exist.
+func bucketID(t *txn, bucket string) (int64, error) {
+	var bucketID int64
+	err := t.QueryRow(`SELECT id FROM buckets WHERE buckets.name = $1`, bucket).Scan(&bucketID)
+	if errors.Is(err, sql.ErrNoRows) {
```
peterjan, line 84: You are ignoring all other errors. For these little helper functions that scan into a field and return an error I personally like named return vars

```
func bucketID(t *txn, bucket string) (bucketID int64, err error) {
    err = t.QueryRow(`SELECT id FROM buckets WHERE buckets.name = $1`, bucket).Scan(&bucketID)
	if errors.Is(err, sql.ErrNoRows) {
        err = s3errs.ErrNoSuchBucket
    }
    return
}

File: sia/persist/sqlite/init.sql [hunk 2]
```diff
@@ -3,6 +3,21 @@
...
+CREATE TABLE buckets (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    created_at INTEGER NOT NULL,
+    name TEXT NOT NULL UNIQUE
+);
+
+CREATE TABLE objects (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    bucket_id INTEGER NOT NULL,
+    name TEXT NOT NULL,
```
peterjan, line 15: I would've gone with key

sia/persist/sqlite/objects.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// DeleteObject deletes the object with the given bucket and name.
+func (s *Store) DeleteObject(accessKeyID, bucket, name string) error {
+	return s.transaction(func(tx *txn) error {
+		bid, err := bucketID(tx, bucket)
+		if err != nil {
+			return err
+		}
+		_, err = tx.Exec("DELETE FROM objects WHERE bucket_id = $1 AND name = $2 ", bid, name)
```
peterjan:
```suggestion
		_, err = tx.Exec("DELETE FROM objects WHERE bucket_id = $1 AND name = $2", bid, name)
```

sia/persist/sqlite/objects.go
```diff
@@ -0,0 +1,75 @@
...
+		return err
+	})
+}
+
+// GetObject retrieves the object with the given bucket and name.
+func (s *Store) GetObject(accessKeyID *string, bucket, name string) (slabs.SealedObject, error) {
+	if accessKeyID == nil {
```
peterjan: it seems very counter intuitive to have ACL in the db layer?

sia/sia_test.go
```diff
@@ -2,30 +2,90 @@ package sia_test
...
-func (s *MemorySDK) Download(ctx context.Context, w io.Writer, obj sdk.Object, opts ...sdk.DownloadOption) error {
-	panic("not implemented")
+func (s *MemorySDK) Download(ctx context.Context, w io.Writer, obj sdk.Object, rnge *s3.ObjectRange) error {
+	uploaded, exists := s.objects[obj.ID()]
+	if !exists {
+		return errors.New("download failed - object not found")
+	}
+	data := uploaded.data
+	if rnge != nil {
+		data = data[rnge.Start : rnge.Start+rnge.Length]
```
peterjan, line 51: I would add boundary checks here to avoid panics

sia/sia.go
```diff
@@ -161,3 +120,43 @@ func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object, uploa
 func (s *Sia) CompleteMultipartUpload(ctx context.Context, accessKeyID, bucket, object, uploadID string, parts []s3.CompletedPart) (*s3.CompleteMultipartUploadResult, error) {
 	return nil, s3errs.ErrNotImplemented
 }
+
+type objectMeta struct {
```
peterjan: Little unit test would be nice

sia/objects.go
```diff
@@ -0,0 +1,164 @@
...
+// contents of the bucket and sort the results into the Contents and
+// CommonPrefixes fields of the returned ObjectsListResult.
+func (s *Sia) ListObjects(ctx context.Context, accessKeyID *string, bucket string, prefix s3.Prefix, page s3.ListObjectsPage) (*s3.ObjectsListResult, error) {
+	return nil, s3errs.ErrNotImplemented
+}
+
+// PutObject puts an object with the given key into the specified bucket.
+func (s *Sia) PutObject(ctx context.Context, accessKeyID string, bucket, object string, r io.Reader, opts s3.PutObjectOptions) (*s3.PutObjectResult, error) {
+	// quick check if the object already exists
+	if _, err := s.store.GetObject(&accessKeyID, bucket, object); !errors.Is(err, s3errs.ErrNoSuchKey) {
```
peterjan: use `HeadObject`?

peterjan (changes requested):

sia/persist/sqlite/objects.go
ChrisSchinnerl: I moved it one layer up for now until we actually have users. We can then decide where to put it.

The downside of not having it in the database code is definitely going to be non atomic verification which might not be the worst, but also another roundtrip since we have to fetch the bucket first to get its permissions.

sia/persist/sqlite/init.sql
ChrisSchinnerl: The reason I didn't go with it was to avoid running into the same issue we had in the past where queries need to put `` around the key since it's also a keyword it seems.

sia/metadata.go
```diff
@@ -0,0 +1,48 @@
...
+	contentMD5 [16]byte
+	meta       map[string]string
+}
+
+func (om *objectMeta) encode() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	enc := types.NewEncoder(buf)
+	_, _ = enc.Write(om.contentMD5[:])
+	enc.WriteUint64(uint64(len(om.meta)))
+	for k, v := range om.meta {
```
peterjan: Map iteration order is non-deterministic in Go, which can result in different encodings for the same metadata map. This makes the encoded output non-deterministic and could complicate debugging or testing. Consider sorting the map keys before encoding to ensure deterministic output.

sia/persist/sqlite/init.sql
peterjan: That's a good reason

peterjan (dismissed):
I would sort the keys before encoding the metadata.

PR #51, opened 2025-11-19
Add `Create|AbortMultipartUpload`

peterjan:


ChrisSchinnerl (commented):
Before we start implementing/merging the Sia backend for multipart uploads, could you enable the compatibility tests for multipart uploads in a separate PR and fix any errors they highlight? That way we know that at least the handlers work as expected before we implement another backend.

sia/persist/sqlite/init.sql
```diff
@@ -22,6 +22,17 @@ CREATE TABLE objects (
     UNIQUE(bucket_id, name)
 );
 
+CREATE TABLE multipart_uploads (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    upload_id BLOB NOT NULL UNIQUE,
+    bucket_id INTEGER NOT NULL,
+    name TEXT NOT NULL,
```
ChrisSchinnerl, line 29: I think both this table and the objects are actually missing a unique composite index on the name and bucket id

ChrisSchinnerl (changes requested):

sia/persist/sqlite/init.sql
peterjan: Well the `object` table has one right, I was on the fence on whether to add one for multipart uploads and eventually decided against it because I don't think we should. I believe S3 lets you happily `CreateMultipartUpload` for the same bucket and key.

ChrisSchinnerl (approved):

PR #53, opened 2025-11-20
Add ListObjects to Sia store

### LINKED ISSUE #46: Implement object listing for Sia backend
Same as the memory backend but within a SQL query

chris124567:
Close #46


sia/persist/sqlite/objects.go
```diff
@@ -68,3 +72,107 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj slabs.SealedObje
...
+			query += ` AND name > ?`
+			args = append(args, *page.Marker)
+		}
+		if prefix.Prefix != "" {
+			query += ` AND name LIKE ?`
+			args = append(args, prefix.Prefix+"%")
+		}
+		query += `
+ORDER BY name
+LIMIT ?;`
```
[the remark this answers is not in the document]
chris124567: this makes code too ugly to be worth it

sia/persist/sqlite/objects.go
```diff
@@ -68,3 +72,107 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj slabs.SealedObje
...
+		return result, nil
+	}
+
+	err := s.transaction(func(tx *txn) error {
+		bid, err := bucketID(tx, bucket)
+		if err != nil {
+			return fmt.Errorf("failed to get bucket ID: %w", err)
+		}
+
+		query := `SELECT name, sia_meta FROM objects WHERE bucket_id = ?`
```
ChrisSchinnerl: This is a good start but I'm afraid it's not that easy. It should be a lot easier than in `renterd` since we no longer allow for sorting by every attribute under the sun but still annoying.

e.g. imagine the following keys
```
a.txt
b.txt
c/sub1.txt
c/sub2.txt
c/sub3.txt
d.txt
```

This is expected to return `a.txt`, `b.txt`, `d.txt` as files and `c/` as a common prefix. 
So the length of the result is 4.
If you set a limit of 4,  your code will not return `d.txt` since the query will not return `d.txt`.

The other issue is that reading all rows under `c/` is a massive performance hit. That folder might contain a million rows but as soon as you encounter the first row starting with `c/` you know that you need to return it as a common prefix and don't need to look at the remaining rows anymore.

I recommend that you run 2 separate queries that you then combine in application code. One query fetches up to `maxKeys` objects, the other up to `maxKeys` common prefixes and finally you merge them.

ChrisSchinnerl (changes requested):

sia/persist/sqlite/objects.go
chris124567: OK I got the case you described to pass:

```
func TestListObjects2(t *testing.T) {
    log := zaptest.NewLogger(t)
    fp := filepath.Join(t.TempDir(), "hostd.sqlite3")

    store, err := OpenDatabase(fp, log)
    if err != nil {
        t.Fatal(err)
    }

    // prepare a bucket
    bucket := "foo"
    if err := store.CreateBucket("", bucket); err != nil {
        t.Fatal(err)
    }

    // upload a few objects
    keys := []string{"a.txt", "b.txt", "c/sub1.txt", "c/sub2.txt", "d.txt"}
    obj := sdk.Object{}
    sealed := obj.Seal(types.GeneratePrivateKey())
    id := obj.ID()

    etag := s3.FormatETag(id[:])
    for _, key := range keys {
        err := store.PutObject("", bucket, key, sealed)
        if err != nil {
            t.Fatal(err)
        }
    }

    ptr := func(s string) *string {
        return &s
    }
    val := func(s *string) string {
        if s == nil {
            return ""
        }
        return *s
    }

    assertCommonPrefixesEqual := func(t *testing.T, expected []string, actual []s3.CommonPrefix) {
        t.Helper()
        if len(expected) != len(actual) {
            t.Fatalf("expected %d common prefixes, got %d", len(expected), len(actual))
        }
        for i := range expected {
            if expected[i] != actual[i].Prefix {
                t.Fatalf("expected common prefix %v, got %v", expected[i], actual[i])
            }
        }
    }

    tests := []struct {
        name       string
        prefix     *string
        delimiter  *string
        marker     *string
        nextMarker *string
        maxKeys    int64

        truncated      bool
        objects        []string
        commonPrefixes []string
    }{
        {
            name:           "All",
            objects:        []string{"a.txt", "b.txt", "d.txt"},
            commonPrefixes: []string{"c/"},
            delimiter:      ptr("/"),
            nextMarker:     ptr("d.txt"),
            maxKeys:        4,
        },
    }
    for _, tc := range tests {
        t.Run(tc.name, func(t *testing.T) {
            resp, err := store.ListObjects(nil, bucket, s3.Prefix{
                Prefix:       val(tc.prefix),
                HasPrefix:    tc.prefix != nil,
                Delimiter:    val(tc.delimiter),
                HasDelimiter: tc.delimiter != nil,
            }, s3.ListObjectsPage{
                Marker:  tc.marker,
                MaxKeys: tc.maxKeys,
            })
            if err != nil {
                t.Fatal(err)
            } else if len(resp.Contents) != len(tc.objects) {
                t.Fatalf("expected %d objects, got %d", len(tc.objects), len(resp.Contents))
            } else if resp.IsTruncated != tc.truncated {
                t.Fatalf("expected truncated=%v, got %v", tc.truncated, resp.IsTruncated)
            }
            for i := range tc.objects {
                if resp.Contents[i].Key != tc.objects[i] {
                    t.Fatalf("expected object %v, got %v", tc.objects[i], resp.Contents[i].Key)
                } else if resp.Contents[i].ETag != etag {
                    t.Fatalf("expected ETag %q, got %q", etag, resp.Contents[i].ETag)
                }
            }
            assertCommonPrefixesEqual(t, tc.commonPrefixes, resp.CommonPrefixes)
            if expectedMarker := val(tc.nextMarker); expectedMarker != resp.NextMarker {
                t.Fatalf("expected marker %v, got %v", expectedMarker, resp.NextMarker)
            }
        })
    }
}
```

sia/persist/sqlite/objects.go
```diff
@@ -68,3 +71,198 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj slabs.SealedObje
 		return err
 	})
 }
+
+// ListObjects lists objects in the specified bucket for the user identified
+// by the given access key. The backend should use the prefix to limit the
```
peterjan, line 119: `accessKeyID` is not used

sia/persist/sqlite/objects.go
```diff
@@ -68,3 +71,198 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj slabs.SealedObje
...
+		args = append(args, *page.Marker)
+	}
+	if prefix.HasPrefix {
+		query += ` AND name LIKE ?`
+		args = append(args, prefix.Prefix+"%")
+	}
+	// Exclude objects that would be common prefixes
+	if prefix.HasDelimiter {
+		query += ` AND name NOT LIKE ?`
+		args = append(args, prefix.Prefix+"%"+prefix.Delimiter+"%")
```
peterjan: @ChrisSchinnerl do we need to support object names with special characters like `"/fileś/śpecial"` I remember we treat the prefix as runes for it to properly handle those weird edge cases

sia/persist/sqlite/objects.go
```diff
@@ -68,3 +71,198 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj slabs.SealedObje
...
+	}
+	defer rows.Close()
+
+	var prefixes []string
+	for rows.Next() {
+		var commonPrefix string
+		if err := rows.Scan(&commonPrefix); err != nil {
+			return nil, fmt.Errorf("failed to scan common prefix: %w", err)
+		}
+		prefixes = append(prefixes, commonPrefix+"/")
```
peterjan: this assumes a `/` delimiter

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/SiaFoundation/s3d/s3"
+	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/sdk"
+	"go.uber.org/zap/zaptest"
+)
+
+func TestListObjects(t *testing.T) {
```
peterjan, line 19: This test is great but I would look at some of the edge cases in `renterd` and copy the ones we want to support over to here. I hope we can drastically simplify it though because that was a mess. This is missing things like `//`, special characters and case sensitivity imo.

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,139 @@
...
+	log := zaptest.NewLogger(t)
+	fp := filepath.Join(t.TempDir(), "hostd.sqlite3")
```
peterjan: 👀

peterjan (changes requested):

sia/persist/sqlite/objects_test.go
ChrisSchinnerl: I'd also take a look at `TestPrefixMatch` in `prefix_test.go`

sia/persist/sqlite/objects.go
ChrisSchinnerl: Yeah. We should. People are definitely going to upload something like that.

sia/persist/sqlite/objects.go
```diff
@@ -68,3 +71,198 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj slabs.SealedObje
...
+
+	if page.Marker != nil && *page.Marker != "" {
+		query += ` AND name > ?`
+		args = append(args, *page.Marker)
+	}
+	if prefix.HasPrefix {
+		query += ` AND name LIKE ?`
+		args = append(args, prefix.Prefix+"%")
+	}
+	// Exclude objects that would be common prefixes
```
ChrisSchinnerl: Comments within functions should be lower case for consistency.

sia/persist/sqlite/objects.go
```diff
@@ -68,3 +71,198 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj slabs.SealedObje
 		return err
 	})
 }
+
+// ListObjects lists objects in the specified bucket for the user identified
+// by the given access key. The backend should use the prefix to limit the
+// contents of the bucket and sort the results into the Contents and
+// CommonPrefixes fields of the returned ObjectsListResult.
+func (s *Store) ListObjects(_ *string, bucket string, prefix s3.Prefix, page s3.ListObjectsPage) (*s3.ObjectsListResult, error) {
+	result := s3.NewObjectsListResult(page.MaxKeys)
```
ChrisSchinnerl: I'd move this into `mergeResults` and return it from there rather than passing a pointer to it.

sia/persist/sqlite/objects_test.go
ChrisSchinnerl, line 1: This is starting to look good. We should add some benchmarking too though. In `renterd` listing objects is ridiculously slow so we should try and avoid that in s3d from the start.

I'd recommend setting it up with a large balanced filesystem hierarchy. e.g. 1000 objects per folder, 10 subfolders per folder and all of that goes a few layers deep. To end up at 1 million objects total or so.

Then you can run multiple benchmarks on that:
1. Listing objects without delimiter
2. Listing the root folder with a delimiter
3. Listing a random bottom folder with a delimiter
4. Listing a random folder with a delimiter

In a perfect world all of these are blazingly fast.

ChrisSchinnerl (changes requested):

sia/persist/sqlite/objects.go
```diff
@@ -68,3 +71,198 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj slabs.SealedObje
...
+	var objects []*s3.Content
+	for rows.Next() {
+		var name string
+		var siaMeta []byte
+		if err := rows.Scan(&name, &siaMeta); err != nil {
+			return nil, fmt.Errorf("failed to scan: %w", err)
+		}
+
+		var obj slabs.SealedObject
+		if err := obj.UnmarshalSia(siaMeta); err != nil {
```
ChrisSchinnerl: Unfortunately it's not that easy. The `obj.ID()` isn't the etag and you can't unseal the `sia_meta` without the SDK.

For now the easiest solution around this is adding a column to `objects` for `last_modified`, `content_md5` and `size` and populate those within `store.PutObject`. 
Then you can ignore `sia_meta` altogether. Which is a large amount of data to load anyway when listing objects.

For that you will need to pass `contentMD5` within `Sia.PutObject` to `store.PutObject`. The other two fields you should be able to get from the sealed object that is already passed in.

chris124567:
```
BenchmarkListObjects/no_delimiter
BenchmarkListObjects/no_delimiter-4                 1448       1169150 ns/op
BenchmarkListObjects/root_delimiter
BenchmarkListObjects/root_delimiter-4                  2     818396310 ns/op
BenchmarkListObjects/folder_bottom_delimiter
BenchmarkListObjects/folder_bottom_delimiter-4                 9     121191774 ns/op
BenchmarkListObjects/folder_delimiter
BenchmarkListObjects/folder_delimiter-4                        9     123950439 ns/op
```

:disappointed: I'll look into full text search on Monday

peterjan:
> ```
> BenchmarkListObjects/no_delimiter
> BenchmarkListObjects/no_delimiter-4                 1448       1169150 ns/op
> BenchmarkListObjects/root_delimiter
> BenchmarkListObjects/root_delimiter-4                  2     818396310 ns/op
> BenchmarkListObjects/folder_bottom_delimiter
> BenchmarkListObjects/folder_bottom_delimiter-4                 9     121191774 ns/op
> BenchmarkListObjects/folder_delimiter
> BenchmarkListObjects/folder_delimiter-4                        9     123950439 ns/op
> ```
> 
> 😞 I'll look into full text search on Monday

Don't forget to `VACUUM ANALYZE` after inserting to rebuild stats tables. This can make a huge difference to the query planner.

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,376 @@
...
+							b.Log(idx)
+						}
+
+						name := strconv.Itoa(idx)
+						layer4 := filepath.Join(layer3, name)
+
+						_, err = tx.Exec(`
+			INSERT INTO objects (bucket_id, name, sia_meta, size, content_md5)
+			VALUES ($1, $2, $3, $4, $5)
+			ON CONFLICT(bucket_id, name) DO UPDATE SET
```
peterjan: should be safe to remove

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,376 @@
...
+		if err != nil {
+			return err
+		}
+
+		var size uint64
+		for _, slab := range sealed.Slabs {
+			size += uint64(slab.Length)
+		}
+
+		for i := 0; i < 1000; i++ {
```
peterjan: `nit` could you turn these into constants and put them at the top of the test so we can easily grasp the scale of the setup

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,376 @@
...
+			}
+			assertCommonPrefixesEqual(t, tc.commonPrefixes, resp.CommonPrefixes)
+			if expectedMarker := val(tc.nextMarker); expectedMarker != resp.NextMarker {
+				t.Fatalf("expected marker %v, got %v", expectedMarker, resp.NextMarker)
+			}
+		})
+	}
+}
+
+func TestListObjectsMatch(t *testing.T) {
```
peterjan, line 212: nice work 👍

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,376 @@
...
+			}, s3.ListObjectsPage{MaxKeys: maxKeys})
+			if err != nil {
+				b.Fatal(err)
+			} else if length := len(resp.Contents) + len(resp.CommonPrefixes); length != maxKeys {
+				b.Fatalf("expected %d objects, got %d", maxKeys, length)
+			}
+		}
+	})
+
+	b.Run("folder_delimiter", func(b *testing.B) {
```
peterjan, line 592: I'm kind of missing a benchmark that lists objects with random prefixes for both no and root delimiter.

chris124567:
```
goos: linux
goarch: amd64
pkg: github.com/SiaFoundation/s3d/sia/persist/sqlite
cpu: Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz
BenchmarkListObjects
BenchmarkListObjects/no_delimiter
BenchmarkListObjects/no_delimiter-4                 1448        808102 ns/op
BenchmarkListObjects/root_delimiter
BenchmarkListObjects/root_delimiter-4                  2     815055824 ns/op
BenchmarkListObjects/random_without_delimiter
BenchmarkListObjects/random_without_delimiter-4                 1383       1011648 ns/op
BenchmarkListObjects/random_with_root_delimiter
BenchmarkListObjects/random_with_root_delimiter-4                  4     267506701 ns/op
BenchmarkListObjects/folder_bottom_delimiter
BenchmarkListObjects/folder_bottom_delimiter-4                     4     269543350 ns/op
BenchmarkListObjects/folder_delimiter
BenchmarkListObjects/folder_delimiter-4                            8     132418215 ns/op
PASS
ok      github.com/SiaFoundation/s3d/sia/persist/sqlite 34.400s
```

OK I will see if there's a way to make delimiters faster

chris124567:
```
BenchmarkListObjects/no_delimiter
BenchmarkListObjects/no_delimiter-4         	       3	 434088475 ns/op
BenchmarkListObjects/root_delimiter
BenchmarkListObjects/root_delimiter-4       	       1	1368323954 ns/op
BenchmarkListObjects/random_without_delimiter
BenchmarkListObjects/random_without_delimiter-4         	       3	 452039091 ns/op
BenchmarkListObjects/random_with_root_delimiter
BenchmarkListObjects/random_with_root_delimiter-4       	     309	   3673739 ns/op
BenchmarkListObjects/folder_bottom_delimiter
BenchmarkListObjects/folder_bottom_delimiter-4          	      44	  25589500 ns/op
BenchmarkListObjects/folder_delimiter
BenchmarkListObjects/folder_delimiter-4                 	       4	 276893845 ns/op
```

I wish this was postgres and I could just use pg_trgm... I think I'm on the right track though

chris124567:
```
BenchmarkListObjects/no_delimiter_no_prefix
BenchmarkListObjects/no_delimiter_no_prefix-4         	    2434	    702068 ns/op
BenchmarkListObjects/root_delimiter
BenchmarkListObjects/root_delimiter-4                 	       2	 909431164 ns/op
BenchmarkListObjects/random_without_delimiter
BenchmarkListObjects/random_without_delimiter-4       	    1798	    851731 ns/op
BenchmarkListObjects/random_with_root_delimiter
BenchmarkListObjects/random_with_root_delimiter-4     	    3153	    400006 ns/op
BenchmarkListObjects/folder_bottom_delimiter
BenchmarkListObjects/folder_bottom_delimiter-4        	    4260	    353755 ns/op
BenchmarkListObjects/folder_delimiter
BenchmarkListObjects/folder_delimiter-4               	     891	   1372094 ns/op
```

Everything is fast except this:
```
	b.Run("root_delimiter", func(b *testing.B) {
		for b.Loop() {
			_, err := store.ListObjects(nil, bucket, s3.Prefix{
				Prefix:       "/",
				HasPrefix:    true,
				Delimiter:    "/1000",
				HasDelimiter: true,
			}, s3.ListObjectsPage{MaxKeys: maxKeys})
			if err != nil {
				b.Fatal(err)
			}
		}
	})
```

:thinking:

sia/persist/sqlite/objects_test.go
```diff
@@ -348,7 +348,9 @@ func BenchmarkListObjects(b *testing.B) {
 	b.Run("root_delimiter", func(b *testing.B) {
 		for b.Loop() {
 			_, err := store.ListObjects(nil, bucket, s3.Prefix{
-				Delimiter:    "/",
+				Prefix:       "/",
+				HasPrefix:    true,
+				Delimiter:    "/1000",
```
peterjan: should this be `/`?

sia/persist/sqlite/objects.go
```diff
@@ -83,3 +88,222 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj *objects.Object)
...
+		bid, err := bucketID(tx, bucket)
+		if err != nil {
+			return fmt.Errorf("failed to get bucket ID: %w", err)
+		}
+
+		// if a string is empty, path.Clean will replace it with "." but the
+		// rest of its [rules](https://pkg.go.dev/path#Clean) are OK to enforce
+		// here
+		if prefix.HasPrefix && prefix.Prefix != "" {
+			prefix.Prefix = strings.ToLower(pathClean(prefix.Prefix))
```
peterjan: We want the prefix to be case sensitive. I missed these cases reviewing your test earlier.

sia/persist/sqlite/objects.go
```diff
@@ -83,3 +88,222 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj *objects.Object)
...
+		}
+		if prefix.HasDelimiter && prefix.Delimiter != "" {
+			prefix.Delimiter = strings.ToLower(pathClean(prefix.Delimiter))
```
peterjan: I think the same goes for delimiters. If I want delimiter `C` and I have an object named `acdc` it shouldn't match that delimiter

chris124567:
@ChrisSchinnerl @peterjan If we restrict the delimiter to ["/"](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html), I think it is possible to optimize this but otherwise it will be difficult.  Are we OK with that?

sia/persist/sqlite/objects.go
```diff
@@ -83,3 +88,222 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj *objects.Object)
...
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
+
+func (s *Store) fetchObjects(tx *txn, bid int64, prefix s3.Prefix, page s3.ListObjectsPage) ([]*s3.Content, error) {
```
peterjan: If you pass a `txn` this shouldn't need the store

sia/persist/sqlite/objects.go
```diff
@@ -83,3 +88,222 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj *objects.Object)
...
+		})
+	}
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("failed to get rows: %w", err)
+	}
+
+	return objs, nil
+}
+
+func (s *Store) fetchCommonPrefixes(tx *txn, bid int64, prefix s3.Prefix, page s3.ListObjectsPage) ([]string, error) {
```
peterjan: If you pass a `txn` this shouldn't need the store

peterjan:
> @ChrisSchinnerl @peterjan If we restrict the delimiter to ["/"](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html), I think it is possible to optimize this but otherwise it will be difficult. Are we OK with that?

Could you elaborate a little on what is slow exactly and why we can optimize for a `/` delimiter but not for instance an `_` delimiter? Often times it's a little difficult to jump in on a PR such as this, for me at least,  because you've done so much of the groundwork already. That's not always immediately apparent from looking at the code and even playing around with the query or benchmarks.

sia/persist/sqlite/init.sql
```diff
@@ -11,16 +12,16 @@ CREATE TABLE buckets (
...
     name TEXT NOT NULL,
     object_id BLOB NOT NULL,
     content_md5 BLOB NOT NULL,
     metadata TEXT NOT NULL,
     size INTEGER NOT NULL,
     updated_at INTEGER NOT NULL,
-    FOREIGN KEY (bucket_id) REFERENCES buckets(id),
     UNIQUE(bucket_id, name)
 );
+CREATE INDEX objects_name ON objects(name);
```
ChrisSchinnerl: Before we try to drop support for arbitrary delimiters could you try one more thing?

Make use of a [clustered index](https://sqlite.org/withoutrowid.html). That means adding `WITHOUT ROWID`, removing `id` and turning `UNIQUE (bucket_id, name)` into `PRIMARY KEY (bucket_id, name)`.

That way the actual data of the table stored on disk will be sorted the way we want to retrieve it already.
Then update the listing code to list all rows from `prefix` until `prefix+0xFF` with a single query and read the rows one by one while adding them to the result rather than using 2 queries and merging the result.

With the composite index this might actually be super fast since listing is now a sequential read from disk even if the number of rows might be quite large.

For the benchmark it should be safe to assume that users will always pick the tightest prefix for a folder.
e.g. if you want to list `/foo/bar/` rclone is going to pick the whole path as the prefix.

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,415 @@
...
+		b.Fatal(err)
+	}
+
+	if _, err := store.db.Exec(`VACUUM;`); err != nil {
+		b.Fatal(err)
+	} else if _, err := store.db.Exec(`ANALYZE;`); err != nil {
+		b.Fatal(err)
+	}
+
+	const maxKeys = 100
```
ChrisSchinnerl: I'd make this 1000 for benchmarking to match the S3 default we have in the API.

sia/persist/sqlite/init.sql
chris124567: I tried this in the most recent commit.  `prefix="/"` with a delimiter is very slow, everything else is very fast.

with maxKeys=100
```
BenchmarkListObjects/no_delimiter_no_prefix
BenchmarkListObjects/no_delimiter_no_prefix-4               2564        448871 ns/op
BenchmarkListObjects/root_delimiter
BenchmarkListObjects/root_delimiter-4                          3     376911106 ns/op
BenchmarkListObjects/random_without_delimiter
BenchmarkListObjects/random_without_delimiter-4             2511        482562 ns/op
BenchmarkListObjects/random_with_root_delimiter
BenchmarkListObjects/random_with_root_delimiter-4           2242        502803 ns/op
BenchmarkListObjects/folder_bottom_delimiter
BenchmarkListObjects/folder_bottom_delimiter-4              9891        127828 ns/op
BenchmarkListObjects/folder_delimiter
BenchmarkListObjects/folder_delimiter-4                      296       4064604 ns/op
PASS
```


with maxKeys=1000:
```
BenchmarkListObjects/no_delimiter_no_prefix
BenchmarkListObjects/no_delimiter_no_prefix-4                314       3736193 ns/op
BenchmarkListObjects/root_delimiter
BenchmarkListObjects/root_delimiter-4                          1    3757831618 ns/op
BenchmarkListObjects/random_without_delimiter
BenchmarkListObjects/random_without_delimiter-4             2553        475163 ns/op
BenchmarkListObjects/random_with_root_delimiter
BenchmarkListObjects/random_with_root_delimiter-4           2386        494856 ns/op
BenchmarkListObjects/folder_bottom_delimiter
BenchmarkListObjects/folder_bottom_delimiter-4              9871        121781 ns/op
BenchmarkListObjects/folder_delimiter
BenchmarkListObjects/folder_delimiter-4                      303       3945087 ns/op
PASS
```

sia/persist/sqlite/migrations.go
```diff
@@ -5,4 +5,11 @@ import "go.uber.org/zap"
 // migrations is a list of functions that are run to migrate the database from
 // one version to the next. Migrations are used to update existing databases to
 // match the schema in init.sql.
-var migrations = []func(tx *txn, log *zap.Logger) error{}
+var migrations = []func(tx *txn, log *zap.Logger) error{
```
peterjan: We don't need migrations (yet) in s3d. You can add it to the schema, add it to `migrations_test.go` and remove this migration.

sia/persist/sqlite/init.sql
ChrisSchinnerl: Could you push that commit for us to take a look and play around with?

chris124567: :+1:

chris124567:
Numbers are looking much better now

with query limit=100:
```
BenchmarkListObjects/no_delimiter_no_prefix
BenchmarkListObjects/no_delimiter_no_prefix-4               2594        454266 ns/op
BenchmarkListObjects/root_delimiter
BenchmarkListObjects/root_delimiter-4                         22      51127609 ns/op
BenchmarkListObjects/random_without_delimiter
BenchmarkListObjects/random_without_delimiter-4             2446        487952 ns/op
BenchmarkListObjects/random_with_root_delimiter
BenchmarkListObjects/random_with_root_delimiter-4           1627        722774 ns/op
BenchmarkListObjects/folder_bottom_delimiter
BenchmarkListObjects/folder_bottom_delimiter-4              7484        157334 ns/op
BenchmarkListObjects/folder_delimiter
BenchmarkListObjects/folder_delimiter-4                     1634        718825 ns/op
PASS
```

with query limit=maxKeys=1000:
```
BenchmarkListObjects/no_delimiter_no_prefix
BenchmarkListObjects/no_delimiter_no_prefix-4                314       3840161 ns/op
BenchmarkListObjects/root_delimiter
BenchmarkListObjects/root_delimiter-4                         19      56778748 ns/op
BenchmarkListObjects/random_without_delimiter
BenchmarkListObjects/random_without_delimiter-4             2323        504158 ns/op
BenchmarkListObjects/random_with_root_delimiter
BenchmarkListObjects/random_with_root_delimiter-4           1549        754312 ns/op
BenchmarkListObjects/folder_bottom_delimiter
BenchmarkListObjects/folder_bottom_delimiter-4              7394        161830 ns/op
BenchmarkListObjects/folder_delimiter
BenchmarkListObjects/folder_delimiter-4                     1656        713605 ns/op
PASS
```

sia/persist/sqlite/objects.go
```diff
@@ -110,3 +113,161 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj *objects.Object)
...
+			}
+
+			if prefix.HasPrefix {
+				query += ` AND o.name >= ? AND o.name < ?`
+				args = append(args, prefix.Prefix, prefix.Prefix+"\xFF")
+			}
+
+			query += ` ORDER BY o.name`
+			query += `  LIMIT ?`
+			args = append(args, 100)
```
ChrisSchinnerl: Let's make this a const.

I'd also add a test case around it to make sure the pagination is correct. 
e.g.
- fetching objects from a directory with more than 100 objects
- fetching objects and common prefixes from a directory where the next marker ends up being in the middle of a subdir

sia/persist/sqlite/init.sql
```diff
@@ -3,24 +3,24 @@
...
     content_md5 BLOB NOT NULL,
     metadata TEXT NOT NULL,
     size INTEGER NOT NULL,
     updated_at INTEGER NOT NULL,
-    FOREIGN KEY (bucket_id) REFERENCES buckets(id),
-    UNIQUE(bucket_id, name)
-);
+    PRIMARY KEY(bucket_id, name)
+) WITHOUT ROWID;
+CREATE INDEX objects_name ON objects(name);
```
ChrisSchinnerl: I think you can drop this index now. But make sure to check if that's true. 
If we can't drop it, the naming should be updated to be consistent with other indices which usually end in `_idx`.

ChrisSchinnerl (changes requested):
This looks pretty good. The benchmark is also great now. Especially for SQLite without any fancy indices.

Since PJ is working on something similar with multipart uploads let's see what he comes up with before merging this. We probably want those approaches to be consistent in their logic.

sia/persist/sqlite/objects.go
```diff
@@ -110,3 +113,161 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj *objects.Object)
...
+		}
+	}
+	return b.String()
+}
+
+// ListObjects lists objects in the specified bucket for the user identified
+// by the given access key. The backend should use the prefix to limit the
+// contents of the bucket and sort the results into the Contents and
+// CommonPrefixes fields of the returned ObjectsListResult.
+func (s *Store) ListObjects(_ *string, bucket string, prefix s3.Prefix, page s3.ListObjectsPage) (result *s3.ObjectsListResult, err error) {
```
peterjan, line 122: @chris124567 can you elaborate a bit on the approach you took here with matching the objects manually and using `maxObjsPerQuery`. I'm assuming that helps the benchmarks? Something I find surprising is that you set `done` to true if you match a common prefix.

I'm also not convinced we need `pathClean`. There's no tests for it either to indicate why it would be necessary we do that. We want to support other delimiters as well, so removing slashes from an object key must cause issues when you list either without or with a different delimiter.

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,488 @@
...
+				assertCommonPrefixesEqual(t, tc.commonPrefixes, resp.CommonPrefixes)
+				if expectedMarker := val(tc.nextMarker); expectedMarker != resp.NextMarker {
+					t.Fatalf("expected marker %v, got %v", expectedMarker, resp.NextMarker)
+				}
+			})
+		}
+	}
+}
+
+func TestListObjectsMatch(t *testing.T) {
```
peterjan, line 212: Even though I appreciate this test and the effort you put in to this already, I feel it's lacking on some fronts. I am not convinced you properly handle UTF8 characters and other delimiters, like `%` for example. In my multipart uploads PR I have a test where I generate a random key space with exotic characters, and then I pick a random delimiter and walk through it and assert I visited every key. That PR still has to go through review but I do feel that's a solid test. Perhaps you could do something similar here?

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,488 @@
...
+
+	"github.com/SiaFoundation/s3d/s3"
+	"github.com/SiaFoundation/s3d/sia/objects"
+	"go.sia.tech/core/types"
+	"go.sia.tech/indexd/sdk"
+	"go.uber.org/zap/zaptest"
+	"lukechampine.com/frand"
+)
+
+func TestListObjects(t *testing.T) {
```
peterjan, line 19: I'd add another set of test cases for a delimiter that's not a `/`. I would also add a section that covers paging through a `/` delimited hierarchy with a `MaxKeys` of 1, to check you have all items and that the markers are correct at all times.

peterjan (changes requested):
It's a little hard to review because the approach you took with list objects is very different than my approach with multipart uploads, and also different with the approach we took in the past. 

What is the reason you can't `SELECT objects UNION ALL SELECT directories`? That should be possible and we should be able to match on the SQL side rather than the Go side. I don't mind it being on the Go side per se but we should have a good reason for it.

sia/persist/sqlite/objects_test.go
peterjan: Scratch that comment on UTF8 chars, I think those are fine.

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,488 @@
...
+	contentMD5 := [16]byte(frand.Bytes(16))
+	etag := s3.FormatETag(contentMD5[:])
+
+	var largeDirectoryKeys []string
+	for i := 0; i < 200; i++ {
+		largeDirectoryKeys = append(largeDirectoryKeys, fmt.Sprintf("large-directory/%d", i))
+	}
+	slices.Sort(largeDirectoryKeys)
+
+	ptr := func(s string) *string {
```
peterjan: What do these helpers accomplish? You pass pointers to have value return the string and if it's `nil` you return an empty string. We can probably get rid of those entirely?

ChrisSchinnerl:
> It's a little hard to review because the approach you took with list objects is very different than my approach with multipart uploads, and also different with the approach we took in the past.
> 
> What is the reason you can't `SELECT objects UNION ALL SELECT directories`? That should be possible and we should be able to match on the SQL side rather than the Go side. I don't mind it being on the Go side per se but we should have a good reason for it.

This is the reason: https://github.com/SiaFoundation/s3d/pull/53#discussion_r2598025263

It's an approach allegedly also used by Minio. It effectively optimizes the data layout on disk to match our query to make optimal use of the index and moves everything that is slow to do in SQL into the application logic. It's also an extremely straightforward approach with good results compared to the overly complicated attempts in `renterd` from the past.

That said `renterd` had the huge disadvantage of requiring aggregate values for size, health etc. Which this approach doesn't really work for since it skips directories using the `xFF` trick. 

My guess is that once we implement the Postgres driver we'll end up having a different solution there using smarter indices but SQLite is quite limited in that regard.

chris124567:
I switched over to PJ's queries and I moved over the the test from his PR.  For some reason the test is failing but I will investigate more tomorrow:

```
objects_test.go:433: expected to visit 10000 uploads, visited 9994
```

(my original queries were missing a lot more than that)

peterjan:
> I switched over to PJ's queries and I moved over the the test from his PR. For some reason the test is failing but I will investigate more tomorrow:

damn man I was going to switch to yours

edit: what do you mean tomorrow 😛

peterjan:
> I switched over to PJ's queries and I moved over the the test from his PR. For some reason the test is failing but I will investigate more tomorrow:

You have to be very mindful when switching because listing multipart uploads is different from objects. This is because of the upload ID marker. You don't have to care about that. That means you can likely get rid of the `PARTITION BY` that gets me the row number and change that by a `GROUP BY` because in your case the object is unique to the bucket. In multipart uploads I can have duplicate keys int he same bucket and have to take into account the upload ID.

chris124567:
> > I switched over to PJ's queries and I moved over the the test from his PR. For some reason the test is failing but I will investigate more tomorrow:
> 
> damn man I was going to switch to yours
> 
> edit: what do you mean tomorrow 😛

By tomorrow I meant today.  I would get "visited 7520" or "visited 6507" objects on my version and I assumed we were going to unify the implementations eventually so I just jumped the gun and did it.

chris124567:
I simplified the query but it turns out the issue was not with the query.  Within each run by pure chance 10-20 of the randomly generated paths would be duplicates of an existing object path.  So instead I just have a set of the keys and only compare the visited count with the number of unique keys.

chris124567:
Need to work on this now... may be missing an index

```
BenchmarkListObjects/no_delimiter_no_prefix
BenchmarkListObjects/no_delimiter_no_prefix-4                288       4129827 ns/op
BenchmarkListObjects/root_delimiter
BenchmarkListObjects/root_delimiter-4                          1    1131510542 ns/op
BenchmarkListObjects/random_without_delimiter
BenchmarkListObjects/random_without_delimiter-4                6     171080468 ns/op
BenchmarkListObjects/random_with_root_delimiter
BenchmarkListObjects/random_with_root_delimiter-4              3     342828884 ns/op
BenchmarkListObjects/folder_bottom_delimiter
BenchmarkListObjects/folder_bottom_delimiter-4                 3     344852836 ns/op
BenchmarkListObjects/folder_delimiter
BenchmarkListObjects/folder_delimiter-4                        3     339899739 ns/op
```

Edit (after 9d4b020d43fff67da94ecbb9e719425e1c1adcb5):
```
BenchmarkListObjects/no_delimiter_no_prefix
BenchmarkListObjects/no_delimiter_no_prefix-4                291       4166480 ns/op
BenchmarkListObjects/root_delimiter
BenchmarkListObjects/root_delimiter-4                          1    1050090947 ns/op
BenchmarkListObjects/random_without_delimiter
BenchmarkListObjects/random_without_delimiter-4             2214        522882 ns/op
BenchmarkListObjects/random_with_root_delimiter
BenchmarkListObjects/random_with_root_delimiter-4           2233        533608 ns/op
BenchmarkListObjects/folder_bottom_delimiter
BenchmarkListObjects/folder_bottom_delimiter-4              3310        363666 ns/op
BenchmarkListObjects/folder_delimiter
BenchmarkListObjects/folder_delimiter-4                      847       1415760 ns/op
```

Not sure there's a solution that is both correct and fast in all cases, at least with SQLite...

chris124567:
With everything passing:

```
BenchmarkListObjects/no_delimiter_no_prefix
BenchmarkListObjects/no_delimiter_no_prefix-4                290       4169392 ns/op
BenchmarkListObjects/root_delimiter
BenchmarkListObjects/root_delimiter-4                         18      56761696 ns/op
BenchmarkListObjects/random_without_delimiter
BenchmarkListObjects/random_without_delimiter-4             2229        555898 ns/op
BenchmarkListObjects/random_with_root_delimiter
BenchmarkListObjects/random_with_root_delimiter-4           1515        767626 ns/op
BenchmarkListObjects/folder_bottom_delimiter
BenchmarkListObjects/folder_bottom_delimiter-4              7856        150414 ns/op
BenchmarkListObjects/folder_delimiter
BenchmarkListObjects/folder_delimiter-4                     1692        690641 ns/op
```

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,607 @@
...
+			}, s3.ListObjectsPage{MaxKeys: maxKeys})
+			if err != nil {
+				b.Fatal(err)
+			} else if (len(result.Contents) + len(result.CommonPrefixes)) == 0 {
+				b.Fatal("no results")
+			}
+		}
+	})
+
+	b.Run("random_without_delimiter", func(b *testing.B) {
```
peterjan, line 546: `nit` the prefix is always at the same level

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,607 @@
...
+			result, err := store.ListObjects(nil, bucket, s3.Prefix{}, s3.ListObjectsPage{MaxKeys: maxKeys})
+			if err != nil {
+				b.Fatal(err)
+			} else if (len(result.Contents) + len(result.CommonPrefixes)) == 0 {
+				b.Fatal("no results")
+			}
+		}
+	})
+
+	b.Run("root_delimiter", func(b *testing.B) {
```
peterjan, line 532: `nit` this is fetching the same objects over and over again, if the result is truncated we could page through the root directory and fetch all objects and prefixes at that level

sia/persist/sqlite/objects_test.go
```diff
@@ -0,0 +1,607 @@
...
+			}, s3.ListObjectsPage{MaxKeys: maxKeys})
+			if err != nil {
+				b.Fatal(err)
+			} else if (len(result.Contents) + len(result.CommonPrefixes)) == 0 {
+				b.Fatal("no results")
+			}
+		}
+	})
+
+	b.Run("random_with_root_delimiter", func(b *testing.B) {
```
peterjan, line 560: `nit` would update the name here `with_root_delimiter` - the delimiter here is a slash, it's not related to listing at the root, especially since in this case we list objects at a prefix 2 levels down

peterjan (approved):
I think the benchmarks could be improved upon a bit but other than that I think this is good to merge.

ChrisSchinnerl (approved):

PR #55, opened 2025-11-21
Update AbortMultipartUpload

peterjan:
I was under the impression that `CompleteMultipartUpload` should be idempotent, but now believe this should not be the case. This contradicts the s3-tests though which has consecutive calls to `CompleteMultipartUpload` that fail with our current implementation. The docs are very clear though, `CompleteMultipartUpload` should return an error if the upload has been aborted or completed.

I do believe we have to update the behaviour of `AbortMultipartUpload`, which is not idempotent but should not fail if you call it twice with the same upload ID. This is because abort frees up the parts resources, and certain parts might still be uploading after the initial call to abort. This PR changes that and adds a small cleanup step to our memory backend to prevent memory leaking.

> As a result, it might be necessary to abort a given multipart upload multiple times in order to completely free all storage consumed by all parts.

straight from the [docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_AbortMultipartUpload.html)

In the [glacier docs](https://docs.aws.amazon.com/cli/latest/reference/glacier/complete-multipart-upload.html) for `CompleteMultipartUpload` it clearly states that it should be idempotent. I propose to keep this implementation and update the s3-tests to remove that consecutive call in the python test suite.

### DIFF
--- a/internal/testutil/backend.go
+++ b/internal/testutil/backend.go
@@ -49,7 +49,6 @@ type (
 
 	object struct {
 		name         string
-		etag         string
 		data         []byte
 		lastModified time.Time
 		metadata     map[string]string
@@ -678,7 +677,6 @@ func (b *MemoryBackend) CompleteMultipartUpload(_ context.Context, accessKeyID, 
 	}
 	bkt.objects[key] = &object{
 		name:         key,
-		etag:         etag,
 		data:         objData,
 		lastModified: time.Now(),
 		metadata:     upload.metadata,
--- a/internal/testutil/backend.go
+++ b/internal/testutil/backend.go
@@ -49,6 +49,7 @@ type (
 
 	object struct {
 		name         string
+		etag         string
 		data         []byte
 		lastModified time.Time
 		metadata     map[string]string
@@ -56,11 +57,13 @@ type (
 	}
 
 	multipartUpload struct {
-		bucket    string
-		key       string
-		metadata  map[string]string
-		parts     map[int]*multipartPart
+		bucket   string
+		key      string
+		metadata map[string]string
+		parts    map[int]*multipartPart
+
 		createdAt time.Time
+		abortedAt time.Time
 	}
 
 	multipartPart struct {
@@ -355,6 +358,13 @@ func (b *MemoryBackend) CreateMultipartUpload(_ context.Context, accessKeyID, bu
 		createdAt: time.Now(),
 	}
 
+	// clean up old aborted uploads
+	for uploadID, upload := range b.multipartUploads {
+		if !upload.abortedAt.IsZero() && time.Since(upload.abortedAt) > time.Hour {
+			delete(b.multipartUploads, uploadID)
+		}
+	}
+
 	return &s3.CreateMultipartUploadResult{
 		UploadID: uploadID,
 	}, nil
@@ -378,6 +388,9 @@ func (b *MemoryBackend) ListMultipartUploads(_ context.Context, accessKeyID, buc
 
 	var entries []entry
 	for id, upload := range b.multipartUploads {
+		if !upload.abortedAt.IsZero() {
+			continue
+		}
 		if upload.bucket != bucket {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


internal/testutil/backend.go
```diff
@@ -509,18 +530,19 @@ func (b *MemoryBackend) UploadPart(_ context.Context, accessKeyID, bucket, key,
...
+
+	// validate uploadID
 	upload, exists := b.multipartUploads[uploadID]
 	if !exists {
 		return nil, s3errs.ErrNoSuchUpload
-	}
-	if upload.bucket != bucket || upload.key != key {
+	} else if upload.bucket != bucket || upload.key != key {
 		return nil, s3errs.ErrNoSuchUpload
 	}
```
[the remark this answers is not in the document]
peterjan: This is actually what we want, even after abort we want to be able to ListParts.

peterjan:
Converted to `DRAFT` for a bit, no need to review just yet. The docs are pretty clear that consecutive calls to complete _can_ return `NoSuchUpload` if the multipart upload was completed already. However, doing so would make the integration tests fail with

```
FAILED s3tests/functional/test_s3.py::test_multipart_upload - botocore.errorfactory.NoSuchUpload: An error occurred (NoSuchUpload) when calling the CompleteMultipartUpload operation: Specified multipart upload does not exist (invalid ID, aborted, or completed).
```

There's also a reference in the [glacier docs](https://docs.aws.amazon.com/cli/latest/reference/glacier/complete-multipart-upload.html) where it does say complete should be idempotent.

Going to fix all other errors first and then revisit this PR.

internal/testutil/backend.go
```diff
@@ -56,11 +56,13 @@ type (
...
-		key       string
-		metadata  map[string]string
-		parts     map[int]*multipartPart
+		bucket   string
+		key      string
+		metadata map[string]string
+		parts    map[int]*multipartPart
+
 		createdAt time.Time
+		abortedAt time.Time
```
ChrisSchinnerl, line 65: Let's treat this the same as `CompleteMultipartUpload` and just return `NoSuchUpload` on multiple calls instead of tracking when the upload was aborted and introducing arbitrary durations for pruning them. According to the API docs that's a valid response.

The Glacier CLI docs might do something different from the API docs such as falling back to a HeadObject in case of a 404 but I'd not use that as an indicator for what the API does.

ChrisSchinnerl (changes requested):

internal/testutil/backend.go
peterjan: I think that's a mistake. 

The glacier docs ref. `CompleteMultipartUpload`, but the regular AWS docs _clearly_ indicate that it might be necessary to abort a given upload multiple times. Straight from the docs

> After a multipart upload is aborted, no additional parts can be uploaded using that upload ID. The storage consumed by any previously uploaded parts will be freed. However, if any part uploads are currently in progress, those part uploads might or might not succeed. As a result, it might be necessary to abort a given multipart upload multiple times in order to completely free all storage consumed by all parts.

I also disagree that `NoSuchUpload` is a valid response for a multipart upload that has already been aborted. If you compare the errors between the docs for `CompleteMultipartUpload` and `AbortMultipartUpload`:

complete:
> Description: The specified multipart upload does not exist. The upload ID might be invalid, or the multipart upload might have been aborted or completed.

abort:
> The specified multipart upload does not exist.

For me this clearly indicates that abort returns `NoSuchUpload` only in the event the ID is total garbage and just never existed.

peterjan: I don't really mind to close this out. I just checked and there's no ceph test that validate this behaviour. I do think it's incorrect though, the docs clearly indicate you should be able to call abort multiple times on the same upload ID. So I don't see why we wouldn't just do that, but up to you.

ChrisSchinnerl: Let's close this out. There is no need to artificially introduce support for nondeterministic behavior that we will never experience in S3d.

peterjan: OK

PR #56, opened 2025-11-21
CopyObject

### LINKED ISSUE #49: Implement CopyObject in Sia backend

ChrisSchinnerl:
Closes https://github.com/SiaFoundation/s3d/issues/49

### DIFF
--- a/s3/s3.go
+++ b/s3/s3.go
@@ -225,9 +225,6 @@ type Backend interface {
 	// - If any referenced part is missing or its ETag does not match,
 	//   [ErrInvalidPart] must be returned.
 	//
-	// - If the list of parts is not strictly ordered by part number,
-	//   [ErrInvalidPartOrder] must be returned.
-	//
 	// NOTE: the parts slice is expected to be sorted by part number and contain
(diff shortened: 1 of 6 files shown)


s3/s3.go
```diff
@@ -225,9 +225,6 @@ type Backend interface {
 	// - If any referenced part is missing or its ETag does not match,
 	//   [ErrInvalidPart] must be returned.
 	//
-	// - If the list of parts is not strictly ordered by part number,
```
peterjan, line 228: why did you remove this? 

From the docs:

```
Special errors
    Error Code: InvalidPartOrder
    Description: The list of parts was not in ascending order. The parts list must be specified in order by part number.
    HTTP Status Code: 400 Bad Request
```

peterjan (changes requested):

s3/s3.go
ChrisSchinnerl: Because you moved validation from the backend to the handler. Meaning that this is no longer true for implementations of the backend interface.

peterjan (approved):

PR #57, opened 2025-11-21
Enable compat tests for CopyObject

ChrisSchinnerl:


s3/s3.go
```diff
@@ -34,7 +34,7 @@ type Backend interface {
 	//
 	// - If the source and destination are the same, the object is kept but its metadata
 	//   is merged with the provided metadata.
-	CopyObject(ctx context.Context, accessKeyID, srcBucket, srcObject, dstBucket, dstObject string, meta map[string]string) (*CopyObjectResult, error)
+	CopyObject(ctx context.Context, accessKeyID, srcBucket, srcObject, dstBucket, dstObject string, replace bool, meta map[string]string) (*CopyObjectResult, error)
```
peterjan, line 38: `nit` feels like it should update the docstring too?

s3/s3.go
```diff
@@ -417,6 +417,7 @@ func (s *s3) routeBase(w http.ResponseWriter, r *http.Request, accessKeyID *stri
 		return
 	}
 	if err != nil {
+		s.logger.Error("failed to handle request", zap.Error(err))
```
peterjan: Wouldn't it be nice to have a logger.With all the various parameters we pass on line :386 and then log failure with that logger here? I understand we can probably deduce it from the logs but still.. we have it handy so we might as well pass all context we have on this line.

s3/objects.go
```diff
@@ -146,11 +146,30 @@ func (s *s3) copyObject(w http.ResponseWriter, r *http.Request, accessKeyID, dst
...
-		return s3errs.ErrNotImplemented // only COPY is supported
+	// copying to the same key without REPLACE is not allowed
+	replace := r.Header.Get("x-amz-metadata-directive") == "REPLACE"
+	if srcBucket == dstBucket && srcObject == dstObject && !replace {
+		return s3errs.ErrInvalidRequest
 	}
 
-	result, err := s.backend.CopyObject(r.Context(), accessKeyID, srcBucket, srcObject, dstBucket, dstObject, meta)
+	// if If-Match or If-None-Match headers are present, handle them
+	ifMatch := r.Header.Get("X-Amz-Copy-Source-If-Match")
```
peterjan, line 156: This probably passes the tox test suite but we should probably add hardening around these headers and whether they're all lower cased or not? I update `writeGetOrHeadObjectHeaders` in my PR too and there I only lowercase the key itself instead of the entire header, which is probably slightly better. 

`.Get("X-Amz-Copy-Source-If-Match")`
`.Get("x-amz-metadata-directive") `

peterjan (changes requested):

s3/objects.go
ChrisSchinnerl: As long as you go through `Get` and `Set`, the headers will be canonicalized. So it really only matters for the user metadata which for some reason has to be lowercase. I updated that to only lowercase the part after the prefix.

peterjan: I think we're missing that in `metadataHeaders` but I added it in #58

peterjan (approved):

PR #61, opened 2025-11-25
Add Get|HeadObjectPart

peterjan:
Both `HeadObject` and `GetObject` have a `PartNumber` parameter that allows querying or fetching information about that part of an object that was uploaded using multipart uploads.  That means we have to keep track of this information for as long as the object exists. This was highlighted by the lack of a `PartsCount` header in the s3 tests, this PR fixes two or three tests from the multipart test suite.

### DIFF
--- a/internal/testutil/backend.go
+++ b/internal/testutil/backend.go
@@ -53,6 +53,13 @@ type (
 		lastModified time.Time
 		metadata     map[string]string
 		contentMD5   [16]byte
+		parts        map[int]objectMultipartPart
+	}
+
+	objectMultipartPart struct {
+		offset     int64
+		length     int64
+		contentMD5 [16]byte
 	}
 
 	multipartUpload struct {
@@ -129,6 +136,7 @@ func (b *MemoryBackend) CopyObject(ctx context.Context, accessKeyID, srcBucket, 
 		lastModified: time.Now(),
 		metadata:     meta,
 		contentMD5:   srcObjct.contentMD5,
+		parts:        srcObjct.parts,
 	}
 	b.buckets[dstBucket].objects[dstObject] = dstObjct
 	return &s3.CopyObjectResult{
@@ -208,8 +216,8 @@ func (b *MemoryBackend) DeleteObjects(ctx context.Context, accessKeyID, bucket s
 }
 
 // GetObject retrieves an object from the specified bucket.
-func (b *MemoryBackend) GetObject(ctx context.Context, accessKeyID *string, bucket, object string, requestedRange *s3.ObjectRangeRequest) (*s3.Object, error) {
-	return b.headOrGetObject(ctx, accessKeyID, bucket, object, requestedRange, false)
+func (b *MemoryBackend) GetObject(ctx context.Context, accessKeyID *string, bucket, object string, requestedRange *s3.ObjectRangeRequest, partNumber *int32) (*s3.Object, error) {
+	return b.headOrGetObject(ctx, accessKeyID, bucket, object, requestedRange, partNumber, false)
 }
 
 // HeadBucket checks if the specified bucket exists and is owned by the user.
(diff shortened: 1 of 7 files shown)


chris124567 (dismissed):

internal/testutil/backend.go
```diff
@@ -48,6 +48,13 @@ type (
 		lastModified time.Time
 		metadata     map[string]string
 		contentMD5   [16]byte
+		parts        map[int]objectMultipartPart
+	}
+
+	objectMultipartPart struct {
+		offset  int64
+		length  int64
+		partMD5 [16]byte
```
ChrisSchinnerl: I'd call this `contentMD5` as well to be consistent.

internal/testutil/testutil.go
```diff
@@ -139,11 +139,12 @@ func (t *S3Tester) DeleteObjects(ctx context.Context, bucket string, objects []s
 }
 
 // GetObject is a convenience wrapper around the AWS SDK's GetObject API.
-func (t *S3Tester) GetObject(ctx context.Context, bucket, object string, rnge *s3.ObjectRangeRequest) (*s3.Object, error) {
+func (t *S3Tester) GetObject(ctx context.Context, bucket, object string, rnge *s3.ObjectRangeRequest, partNumber *int32) (*s3.Object, error) {
```
ChrisSchinnerl: I'd add another `GetObjectPart` for this since a range and a part are mutually exclusive.

ChrisSchinnerl (changes requested):

internal/testutil/testutil.go
peterjan: Is that really necessary? If you look at `service.GetObjectInput` it takes a `PartNumber` so I'm not sure if this should be split. It's mutually inclusive in the sense that the partNumber is ignored if a range is given, but since `GetObjectInput` takes both I don't feel it's enough reason to split these up?

ChrisSchinnerl: `GetObjectInput` takes both because the AWS SDK is a horrible piece of Go code which was not hand-written by a Go dev but generated by a tool. The point of having these methods on `S3Tester` is to provide a nice and safer to use abstraction for our tests not to propagate the nastiness of that SDK into our actual tests.

internal/testutil/backend.go
```diff
@@ -697,7 +714,37 @@ func (b *MemoryBackend) headOrGetObject(_ context.Context, accessKeyID *string,
...
 		return nil, err
+	} else if rnge != nil {
+		partNumber = nil // ignore part number when a range is requested
 	}
+
+	var partCount *int32
+	var contentMD5 [16]byte
+	if partNumber != nil {
+		if len(obj.parts) == 0 {
+			if *partNumber != 1 {
```
ChrisSchinnerl, line 725: Why would that be the case? After all you need to provide [at least 1 part](https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html) when completing the upload so `len(obj.parts) == 0` would only indicate that the upload isn't a multipart upload in which case you can`t reference the part number?

peterjan: The rule is that for non multipart uploads you can ask for `partNumber=1` and you get the full object.
Considering the ceph test suite has a dedicated test for this edge case, called `test_non_multipart_get_part`, I was inclined to support it.  

The docs have two mentions of `partNumber`, both don't explicitly prevent that behaviour. 

1 [partNumber](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax)
Part number of the object being read. This is a positive integer between 1 and 10,000. Effectively performs a 'ranged' GET request for the part specified. Useful for downloading just a part of an object.

2 [x-amz-mp-parts-count](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_ResponseSyntax)
The count of parts this object has. This value is only returned if you specify partNumber in your request and the object was uploaded as a multipart upload.

 I'm assuming they added this little hack to get/head object to facilitate composing multipart uploads from already uploaded objects, regardless of whether it was a multipart upload. The docs clearly don't prevent it and the fact that `partNumber` is specified on an "object" scope and not explicitly restricted to a multipart upload makes me believe it's just another quirk

ChrisSchinnerl (approved):

PR #62, opened 2025-11-25
Add UploadPartCopy

peterjan:
This PR adds `UploadPartCopy`, this should unlock some more of the s3-tests. After merging this we should be close to a green test suite on multipart uploads.

`NOTE`: there's a couple of headers we currently do not support, `if-none-match` and `if-unmodified-since` stuff. My plan is to tackle those in a F/U, or not if we decide not to support those. I'm probably going to copy whatever we support/do on `CopyObject`


chris124567 (dismissed):

s3/multipart.go
```diff
@@ -482,6 +553,51 @@ func FormatMultipartETag(hash []byte, partCount int) string {
...
+	}
+
+	start, err := strconv.ParseInt(parts[0], 10, 64)
+	if err != nil || start < 0 {
+		return ObjectRange{}, s3errs.ErrInvalidArgument
+	}
+	end, err := strconv.ParseInt(parts[1], 10, 64)
+	if err != nil || end < start {
+		return ObjectRange{}, s3errs.ErrInvalidArgument
+	}
```
ChrisSchinnerl: You can use `fmt.Sscanf("bytes=%d-%d", &start, &end)` to make this a lot shorter since a failure always means `ErrInvalidArgument`

ChrisSchinnerl (changes requested):
Looking pretty good but got conflicts

ChrisSchinnerl (approved):

PR #64, opened 2025-11-26
Update CompleteMultipartUpload

peterjan:
We need to be able to handle completed parts with the same part number and keep the last one.


internal/testutil/backend.go
```diff
@@ -607,16 +607,25 @@ func (b *MemoryBackend) CompleteMultipartUpload(_ context.Context, accessKeyID,
...
+		if i > 0 && completed.PartNumber < prev {
 			return nil, s3errs.ErrInvalidPartOrder
 		}
 		prev = completed.PartNumber
+		if len(deduped) > 0 && deduped[len(deduped)-1].PartNumber == completed.PartNumber {
+			deduped[len(deduped)-1] = completed // overwrite duplicate
+		} else {
+			deduped = append(deduped, completed)
+		}
+	}
```
ChrisSchinnerl: All of this can be moved into the handler itself so that every backend benefits from it. The interface should be able to assume that the input is already ordered and deduplicated.

ChrisSchinnerl (changes requested):

s3/s3.go
```diff
@@ -227,6 +227,9 @@ type Backend interface {
 	//
 	// - If the list of parts is not strictly ordered by part number,
```
ChrisSchinnerl, line 228: This is now no longer necessary.

ChrisSchinnerl (approved):

PR #69, opened 2025-12-03
Fix `parseRange`

peterjan:
Forgot to run the test suite after updating `parseRange` - this fixes the implementation.

### DIFF
--- a/s3/multipart.go
+++ b/s3/multipart.go
@@ -3,7 +3,9 @@ package s3
 import (
 	"encoding/base64"
 	"encoding/hex"
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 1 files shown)


s3/multipart.go
ChrisSchinnerl, line 1: What was the regression?

peterjan: Well I had a working impl. and change it last minute without rerunning the test suite. The updated version that uses `fmt.Sscanf` failed various test cases in the ceph test suite, `test_multipart_copy_improper_range` for one but there were 5 or 6 I think.

ChrisSchinnerl: What was the exact cause though? At a first glance it looks like you missed the `size <= 0` check and `end >= size` one but the middle part should be equivalent to `Sscanf`.

peterjan: Can't pass multiple ranges for one. E.g. `bytes=0-2,3-5` should be an invalid argument but `Sscanf` would parse our 0,2 and ignore the rest. Do you want me to rewrite it using `Sscanf`? I rewrote it with an extra `suffix` parameter that would grab anything after `bytes=%d-%d%s` but ran into other tests that failed.

peterjan: I was missing the `io.EOF` check when I tried extracting the `suffix`.

chris124567 (dismissed):

s3/multipart.go
```diff
@@ -568,17 +570,23 @@ func FormatMultipartETag(hash []byte, partCount int) string {
...
+	} else if header == "" {
 		return ObjectRange{Start: 0, Length: size}, nil
 	}
 
 	var start, end int64
-	_, err := fmt.Sscanf(header, "bytes=%d-%d", &start, &end)
-	if err != nil {
+	var suffix string
+	_, err := fmt.Sscanf(header, "bytes=%d-%d%s", &start, &end, &suffix)
+	if err != nil && !errors.Is(err, io.EOF) {
```
ChrisSchinnerl:
```suggestion
	n, _ := fmt.Sscanf(header, "bytes=%d-%d%s", &start, &end, &suffix)
	if n != 2 {
```

This makes sure that exactly `start` and `end` were read successfully and saves you from checking the error

s3/multipart.go
```diff
@@ -568,17 +570,23 @@ func FormatMultipartETag(hash []byte, partCount int) string {
...
 		return ObjectRange{}, s3errs.ErrInvalidArgument
 	}
 
-	if start < 0 || end < start {
+	if suffix != "" {
```
ChrisSchinnerl: No need to check the suffix if `n == 2`

ChrisSchinnerl (changes requested):

s3/multipart.go
peterjan: We have to check the error because `"bytes=5-6"` will return `io.EOF` too.

We can do the following but imo that is worse. This is why I avoided using `fmt.Sscanf` in the first place actually and went for the more verbose prefix splits and `strconv.ParseInt`
```
    var start, end int64
	var suffix string
	n, err := fmt.Sscanf(header, "bytes=%d-%d%s", &start, &end, &suffix)
	if (err != nil && !errors.Is(err, io.EOF)) || n != 2 {
		return ObjectRange{}, s3errs.ErrInvalidArgument
	}
	```

peterjan: Oh I'm sorry. I missed the fact that you were ignoring the error. I prefer checking the error over checking `n` but don't mind updating it if you feel differently.

ChrisSchinnerl (approved):

PR #71, opened 2025-12-03
DeleteObjects

ChrisSchinnerl:
This PR implements `DeleteObjects` for the Sia backend and enables the object deletion related compatibility tests. 

Requires https://github.com/SiaFoundation/s3-tests/pull/3

### DIFF
--- a/internal/testutil/backend.go
+++ b/internal/testutil/backend.go
@@ -189,7 +189,7 @@ func (b *MemoryBackend) DeleteObject(ctx context.Context, accessKeyID, bucket st
 	if o, exists := bkt.objects[object.Key]; exists &&
 		((object.ETag != nil && s3.FormatETag(o.contentMD5[:]) != *object.ETag) ||
 			(object.Size != nil && int64(len(o.data)) != *object.Size) ||
-			(object.LastModifiedTime != nil && !o.lastModified.Round(time.Second).Equal(object.LastModifiedTime.StdTime()))) {
+			(object.LastModifiedTime != nil && !o.lastModified.Truncate(time.Second).Equal(object.LastModifiedTime.StdTime()))) {
 		return nil, s3errs.ErrPreconditionFailed
 	}
 	delete(bkt.objects, object.Key)
@@ -213,7 +213,7 @@ func (b *MemoryBackend) DeleteObjects(ctx context.Context, accessKeyID, bucket s
 		if exists &&
 			((obj.ETag != nil && s3.FormatETag(o.contentMD5[:]) != *obj.ETag) ||
 				(obj.Size != nil && int64(len(o.data)) != *obj.Size) ||
-				(obj.LastModifiedTime != nil && !o.lastModified.Round(time.Second).Equal(obj.LastModifiedTime.StdTime()))) {
+				(obj.LastModifiedTime != nil && !o.lastModified.Truncate(time.Second).Equal(obj.LastModifiedTime.StdTime()))) {
 			res.Error = append(res.Error, s3.ErrorResult{
 				Key:     obj.Key,
 				Code:    s3errs.ErrPreconditionFailed.Code,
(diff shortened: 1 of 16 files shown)


sia/persist/sqlite/objects.go
```diff
@@ -2,21 +2,49 @@ package sqlite
...
+			var updatedAt time.Time
+			err = tx.QueryRow("SELECT content_md5, size, updated_at FROM objects WHERE bucket_id = $1 AND name = $2", bid, objectID.Key).
+				Scan((*sqlMD5)(&contentMD5), &size, (*sqlTime)(&updatedAt))
+			if errors.Is(err, sql.ErrNoRows) {
+				return nil // object doesn't exist, nothing to delete
+			} else if err != nil {
+				return err
+			}
+
+			if objectID.ETag != nil && *objectID.ETag != hex.EncodeToString(contentMD5[:]) {
```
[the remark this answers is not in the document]
ChrisSchinnerl: It seems to be fine according to the compat tests but I added a trim for safety.

internal/testutil/backend.go
```diff
@@ -179,17 +179,20 @@ func (b *MemoryBackend) DeleteBucket(ctx context.Context, accessKeyID, name stri
...
 	bkt, exists := b.buckets[bucket]
 	if !exists {
 		return nil, s3errs.ErrNoSuchBucket
 	} else if bkt.owner != b.accessKeys[accessKeyID].owner {
 		return nil, s3errs.ErrAccessDenied
 	}
-	if _, exists := bkt.objects[object]; !exists {
-		return nil, s3errs.ErrNoSuchKey
+	if o, exists := bkt.objects[object.Key]; exists &&
+		((object.ETag != nil && s3.FormatETag(o.contentMD5[:]) != *object.ETag) ||
```
peterjan: `nit` Can we add a method to our `object` struct to clean this up? This is quite verbose and really not that important in a way.  Something like `exists && !o.Match(object)`

sia/persist/sqlite/objects.go
```diff
@@ -2,21 +2,49 @@ package sqlite
...
+				return s3errs.ErrPreconditionFailed
+			}
+			if objectID.Size != nil && *objectID.Size != size {
+				return s3errs.ErrPreconditionFailed
+			}
+			if objectID.LastModifiedTime != nil && !updatedAt.Truncate(time.Second).Equal(objectID.LastModifiedTime.StdTime()) {
```
peterjan, line 41: This feels a little fishy? Why isn't it an exact match? Even if `X-Amz-If-Match-Last-Modified-Time` is truncated to a second shouldn't we make sure that's reflected in our database and store it in a way it can be matched straight up?

sia/objects_test.go
```diff
@@ -294,3 +295,78 @@ func TestCopyObject(t *testing.T) {
 		t.Fatalf("last modified mismatch: expected after %v, got %v", copyTime, obj.LastModified)
 	}
 }
+
+func TestDeleteObjects(t *testing.T) {
```
peterjan, line 299: There's currently no testing around the precondition failed errors.

peterjan (changes requested):

internal/testutil/testutil.go
```diff
@@ -123,19 +123,24 @@ func (t *S3Tester) DeleteObject(ctx context.Context, bucket, object string) erro
 	return err
 }
 
-// DeleteObjects deletes multiple S3 objects at once. If quiet is set to true,
-// the response will only contain errors.
-func (t *S3Tester) DeleteObjects(ctx context.Context, bucket string, objects []string, quiet *bool) (*service.DeleteObjectsOutput, error) {
+// ObjectIdentifier creates an ObjectIdentifier for the given key.
```
peterjan: `nit` `ObjectIdentifiers`

peterjan (approved):

PR #72, opened 2025-12-03
Add `UploadPart`

peterjan:
This PR adds `UploadPart`, the parts are cached on disk before we'll upload them in one go when the multipart upload is completed. The most important things here is that more than one multipart upload can be created for the same bucket/object combination. Parts can be retried and should be overwritten for the same part number. We can't be fully ACID between our on-disk cache and the database, so we use the database as a single source of truth and will cleanup files on disk on startup and filter parts without a proper `md5` checksum when the upload gets completed.


sia/multipart.go
```diff
@@ -48,8 +60,104 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
...
+	var contentSHA256 *[32]byte
+	if opts.ContentSHA256 != nil {
+		contentSHA256 = new([32]byte)
+		copy(contentSHA256[:], sha256Hash.Sum(nil))
+		if *opts.ContentSHA256 != *contentSHA256 {
+			return nil, s3errs.ErrBadDigest
+		}
+	}
+
+	// TODO: If a part already exists, we could preserve it by moving the current
```
peterjan: What do you think of moving an existing part if necessary and roll back on error? It's a little gnarly but if we do that one more thing we cover a lot of the crash edge cases. I still have to cleanup on startup.

sia/multipart_test.go
```diff
@@ -92,3 +97,56 @@ func TestAbortMultipartUpload(t *testing.T) {
 	err = s3Tester.AbortMultipartUpload(t.Context(), bucket, object, uploadID)
 	testutil.AssertS3Error(t, s3errs.ErrNoSuchUpload, err)
 }
+
+func TestMultipartAddPart(t *testing.T) {
```
peterjan, line 101: This test is far from being done, I need some store methods though to list parts and multi part uploads to assert what has made it in the store. I'm going to extend this test after add the list methods.

sia/multipart.go
```diff
@@ -48,8 +60,104 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
...
+func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object, uploadID string, r io.Reader, opts s3.UploadPartOptions) (_ *s3.UploadPartResult, err error) {
+	// quick check if the bucket exists
+	if err := s.store.HeadBucket(accessKeyID, bucket); err != nil {
+		return nil, err
+	}
+
+	// add part metadata to the database
+	if err := s.store.AddMultipartPart(bucket, object, uploadID, opts.PartNumber); err != nil {
+		return nil, fmt.Errorf("failed to add multipart part: %w", err)
+	}
```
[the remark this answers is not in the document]
peterjan: Yes incomplete parts will be filtered. I'll document it later

sia/multipart.go
```diff
@@ -48,8 +60,104 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
 }
 
 // UploadPart uploads a single multipart part.
-func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object, uploadID string, r io.Reader, opts s3.UploadPartOptions) (*s3.UploadPartResult, error) {
-	return nil, s3errs.ErrNotImplemented
+func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object, uploadID string, r io.Reader, opts s3.UploadPartOptions) (_ *s3.UploadPartResult, err error) {
```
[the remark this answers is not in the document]
peterjan: I ignored you once, I'll ignore you twice

sia/persist/sqlite/types.go
```diff
@@ -12,6 +12,7 @@ import (
 var (
 	_ scannerValuer = (*sqlHash256)(nil)
 	_ scannerValuer = (*sqlMD5)(nil)
+	_ scannerValuer = (*sqlSHA256)(nil)
```
ChrisSchinnerl: This is redundant since we already have `sqlHash256` for generic `32` byte hashes.

sia/multipart.go
```diff
@@ -2,13 +2,25 @@ package sia
...
 	"io"
+	"os"
+	"path/filepath"
 
 	"github.com/SiaFoundation/s3d/s3"
 	"github.com/SiaFoundation/s3d/s3/s3errs"
+	"lukechampine.com/frand"
 )
 
+func randUUID() (uuid [8]byte) {
```
ChrisSchinnerl: (nit) `randUploadID` considering this has nothing to do with `UUID`s

sia/multipart.go
```diff
@@ -48,8 +60,104 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create part file: %w", err)
+	}
+	defer partFileTmp.Close()
+
+	// defer a best effort cleanup on error
+	defer func() {
+		if err != nil {
+			_ = os.Remove(partPathTmp)
+			_ = os.Remove(partPathFinal)
```
ChrisSchinnerl: This current approach is not very thread safe. If an upload fails you might end up blindly deleting a good part.

sia/multipart.go
```diff
@@ -48,8 +60,104 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
...
+	writers := []io.Writer{partFileTmp, md5Hash}
+
+	var sha256Hash hash.Hash
+	if opts.ContentSHA256 != nil {
+		sha256Hash = sha256.New()
+		writers = append(writers, sha256Hash)
+	}
+
+	// copy data and validate size
+	buf := make([]byte, 64*1024) // 64 KiB buffer for efficient copying
```
ChrisSchinnerl: You can use `io.Copy` since it's already buffered unless 64kib somehow performs better than 32kib.

sia/persist/sqlite/init.sql
```diff
@@ -33,6 +33,18 @@ CREATE TABLE multipart_uploads (
 );
 CREATE INDEX multipart_uploads_bucket_id_name_idx ON multipart_uploads(bucket_id, name);
 
+CREATE TABLE multipart_parts (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    multipart_upload_id INTEGER NOT NULL,
+    part_number INTEGER NOT NULL,
+    content_md5 BLOB,
```
ChrisSchinnerl: I thought a bit more about this and the requirement of uploading multiple parts in parallel and I think we can simplify this by adding a `filename` column to `multipart_parts`.

The total location of the part on disk would look like this:
```
location := "./uploads/<upload_id>/<part_num>/<random>.part"
```

Then we can perform the full upload without any database interaction for any concurrent number of parts. When the upload is done, we call `AddMultipartPart` with the `<random>` part which upserts a row.

That way
1. we can atomically swap out successful parts without a rename and reduce roundtrips to 1
2. we don't need any rollback mechanism since we only update the row upon success

`AddMultipartPart` can return the existing file location for us to clean up immediately for convenience. 

I would also not worry about cleanups in case of crashes for now. The whole directory will be cleaned up once `AbortMultipartUpload` is called and s3d shouldn't crash often enough for this to make a difference.

ChrisSchinnerl (changes requested):

sia/persist/sqlite/init.sql
peterjan: Moving the part_number into the path as a folder was a great suggestion 👍 thanks.

sia/persist/sqlite/store.go
```diff
@@ -106,27 +106,39 @@ func doTransaction(db *sql.DB, log *zap.Logger, fn func(tx *txn) error) error {
...
+		return nil, err
+	}
 
-	db, err := sql.Open("sqlite3", sqliteFilepath(fp))
+	sqliteVersion, _, _ := sqlite3.Version()
+	log.Debug("database initialized", zap.String("sqliteVersion", sqliteVersion), zap.Int("schemaVersion", len(migrations)+1), zap.String("path", path))
+	return store, nil
+}
+
+func initDB(path string) (*sql.DB, error) {
```
ChrisSchinnerl: You can avoid the changes in this file if you change `initTestDB` to

```go
func initTestDB(t testing.TB, log *zap.Logger) *testStore {
	store, err := OpenDatabase(filepath.Join(t.TempDir(), "s3d.sqlite"), log)
	if err != nil {
		t.Fatal(err)
	}
	t.Cleanup(func() {
		if err := store.Close(); err != nil {
			panic(err)
		}
	})

	return &testStore{
		Store: store,
		db:    store.db,
		t:     t,
	}
}
```

sia/multipart.go
```diff
@@ -48,8 +61,100 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
...
-	return nil, s3errs.ErrNotImplemented
+func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object, uploadID string, r io.Reader, opts s3.UploadPartOptions) (_ *s3.UploadPartResult, err error) {
+	// quick check if the bucket exists
+	if err := s.store.HeadBucket(accessKeyID, bucket); err != nil {
+		return nil, err
+	}
+
+	// create part directory
+	partDir := filepath.Join(s.directory, uploadID, fmt.Sprintf("%d", opts.PartNumber))
+	if err := os.MkdirAll(partDir, 0700); err != nil {
```
ChrisSchinnerl: `0600` should be plenty?

Also I think this might cause a race. If we delete the upload directory in `CompleteMultipartUpload` and this effectively recreates it. So at the very least we probably want to use `Mkdir` here rather than `MkdirAll` to prevent the creation of part directories for uploads that don't exist.

sia/multipart.go
```diff
@@ -48,8 +61,100 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
 }
 
 // UploadPart uploads a single multipart part.
-func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object, uploadID string, r io.Reader, opts s3.UploadPartOptions) (*s3.UploadPartResult, error) {
-	return nil, s3errs.ErrNotImplemented
+func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object, uploadID string, r io.Reader, opts s3.UploadPartOptions) (_ *s3.UploadPartResult, err error) {
+	// quick check if the bucket exists
+	if err := s.store.HeadBucket(accessKeyID, bucket); err != nil {
```
ChrisSchinnerl: Check if the Multipart Upload exists and not just the bucket

sia/multipart.go
```diff
@@ -48,8 +61,100 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
...
+			return nil, s3errs.ErrBadDigest
+		}
+	}
+
+	// sync part file
+	if err := partFile.Sync(); err != nil {
+		return nil, fmt.Errorf("failed to sync part file: %w", err)
+	}
+
+	// sync parent directory
```
ChrisSchinnerl, line 163: Don't quote me on this but I don't think you need to sync the dir here. Syncing the file means it's stored on disk and it would be odd if the file existed but the dir didn't.

sia/multipart.go
```diff
@@ -48,8 +61,100 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
...
+	dir, err := os.Open(partDir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open multipart tmp dir: %w", err)
+	} else if err := errors.Join(dir.Sync(), dir.Close()); err != nil {
+		return nil, fmt.Errorf("failed to sync multipart part file and dir: %w", err)
+	}
+
+	// get relative location
+	location, err := filepath.Rel(s.directory, partPath)
```
ChrisSchinnerl: I'd pass the filename to `AddMultipartPart` instead of a full relative location.
We always know the upload id and part id when looking up parts and that's all we need to derive the path.

sia/persist/sqlite/store.go
peterjan: lol

sia/multipart.go
peterjan: It depends on how we want to approach this. In the past we've always looked at this from the perspective of crash safety. Syncing the parent directory as well as syncing the file is definitely the most correct and safe approach. 

From the [docs](https://www.man7.org/linux/man-pages/man2/fsync.2.html)
> “Calling fsync() does not necessarily ensure that the entry in the directory containing the file has also reached disk. For that an explicit fsync() on a file descriptor for the directory is also needed.”

I would probably keep it? It doesn't hurt us, syncing the file will be faster but we don't really care in this case imo.

sia/multipart.go
peterjan: I considered this actually but kind of liked it contained the full relative location so we can be lazy in `CompleteMultipartUpload`. I'll change it though it's redundant information after all.

sia/multipart.go
peterjan: `0700` is what we've always used? I think 0700 is the default for directories and 0600 for files. I'll change it to `Mkdir`, I guess I didn't reconsider this line when changing the path layout.

sia/multipart.go
peterjan: We used to have a two setup part update in the database which did that... thanks for catching that

sia/persist/sqlite/init.sql
```diff
@@ -33,6 +33,19 @@ CREATE TABLE multipart_uploads (
 );
 CREATE INDEX multipart_uploads_bucket_id_name_idx ON multipart_uploads(bucket_id, name);
 
+CREATE TABLE multipart_parts (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    multipart_upload_id INTEGER NOT NULL,
+    part_number INTEGER NOT NULL,
+    filename TEXT NOT NULL,
```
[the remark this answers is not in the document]
peterjan: not even true?

sia/multipart.go
```diff
@@ -44,12 +62,114 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
...
+	// check if the multipart upload exists
+	if err := s.store.HasMultipartUpload(bucket, object, uploadID); err != nil {
+		return nil, err
+	}
+
+	// create part directory
+	partDir := filepath.Join(s.directory, uploadID, fmt.Sprintf("%d", opts.PartNumber))
+	if err := os.Mkdir(partDir, 0700); errors.Is(err, os.ErrNotExist) {
+		return nil, s3errs.ErrNoSuchUpload
+	} else if err != nil {
```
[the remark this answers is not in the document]
peterjan: This is the behaviour we want.

sia/multipart.go
```diff
@@ -22,6 +35,11 @@ func (s *Sia) CreateMultipartUpload(ctx context.Context, accessKeyID, bucket, ob
 		return nil, fmt.Errorf("failed to create multipart upload: %w", err)
 	}
 
+	// create multipart upload directory
+	if err := os.Mkdir(filepath.Join(s.directory, uploadID), 0700); err != nil {
```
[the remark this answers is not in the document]
peterjan: I didn't bother but I guess we probably should do that 👍

sia/persist/sqlite/multipart.go
```diff
@@ -53,28 +44,75 @@ func (s *Store) CreateMultipartUpload(bucket, name string, meta map[string]strin
...
+		if err != nil {
+			return err
+		}
+
+		var sha256Value any
+		if contentSHA256 != nil {
+			sha256Value = sqlHash256(*contentSHA256)
+		}
+
+		var prev sql.NullString
```
ChrisSchinnerl: This can never be `NULL`

sia/persist/sqlite/multipart.go
```diff
@@ -53,28 +44,75 @@ func (s *Store) CreateMultipartUpload(bucket, name string, meta map[string]strin
...
+		err = tx.QueryRow(`SELECT filename FROM multipart_parts WHERE multipart_upload_id = $1 AND part_number = $2`, uid, partNumber).Scan(&prev)
+		if err != nil && err != sql.ErrNoRows {
```
ChrisSchinnerl: `!errors.Is(err, sql.ErrNoRows)`

sia/multipart.go
```diff
@@ -22,6 +36,18 @@ func (s *Sia) CreateMultipartUpload(ctx context.Context, accessKeyID, bucket, ob
...
```
ChrisSchinnerl: Considering that the database is our source of truth regarding the existence of an upload I'd switch this around. Create the folder first and then follow it up with `CreateMultipartUpload`.

A folder on disk but not in the database is a lot less harmful than the other way round. Because we list uploads using the database and not the folders. So when it exists in the database we expect it to exist on disk too.

sia/multipart.go
```diff
@@ -48,8 +82,113 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
 }
 
 // UploadPart uploads a single multipart part.
-func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object, uploadID string, r io.Reader, opts s3.UploadPartOptions) (*s3.UploadPartResult, error) {
-	return nil, s3errs.ErrNotImplemented
+func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object, uploadID string, r io.Reader, opts s3.UploadPartOptions) (_ *s3.UploadPartResult, err error) {
+	// check auth and bucket existence
+	if err := s.store.HeadBucket(accessKeyID, bucket); err != nil {
```
ChrisSchinnerl: This is redundant now. `HasMultipartUpload` will return the appropriate error if the bucket doesn't exist.

sia/multipart.go
```diff
@@ -39,6 +65,14 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
 		return err
 	}
 
+	// remove multipart upload directory
```
ChrisSchinnerl, line 71: Same thing as before. This should happen in reverse order of `CreateMultipartUpload`.
First delete from the database and then from disk. 

It's going to be easy enough to follow up with some cleanup routine that deletes folders that don't belong to a database entry.

ChrisSchinnerl (changes requested):
Got a few more potential leftovers but looks good otherwise.

sia/multipart.go
peterjan: No we need the auth check on the bucket owner. Auth checks is something we should tackle by the way, I think I asked you about that before once. We pass `accessKeyID` around to the store for no reason at times. I'd rather open an issue to fix our auth handling than update this because we do this everywhere.

sia/multipart.go
peterjan: The directory contains the upload ID, which is generated by `CreateMultipartUpload`. We could change that but I don't think we should.

sia/multipart.go
peterjan: I don't mind reversing this one but I did think about the order. I figured that because we don't error out when the disk delete fails, it makes more sense to do that first so if you call it multiple times we'll always be trying to delete both disk and database entry. I guess it does make more sense to revert it though.

sia/multipart.go
ChrisSchinnerl: `HeadBucket` doesn't perform an auth check and we don't have owners?

I think once we actually need auth checks we'll not perform them in the backend but in the S3 handlers instead. e.g. by adding a `CheckPermissions(userID string, perms ...Permission)` function to the `backend` interface. Where a `Permission` is a combination of bucket and some flags.

e.g. the CopyObject handler knows that it needs read permissions on the source bucket and write permissions on the destination bucket. So it can verify that before calling the backend.

sia/multipart.go
ChrisSchinnerl: Yeah that's fine. I'm just saying create the folder and then update the database rather than the other way round. The directory should still contain the upload id.

sia/persist/sqlite/multipart.go
peterjan: Right you are

sia/multipart.go
peterjan: I have to create the upload in the database, that gives me the ID. I can then use that ID to create the directory on disk. It doesn't work the other way around right. 

edit: I'll update it so the upload ID is generated and we pass it to the store. Not sure why but that just didn't feel right to me.

sia/multipart.go
peterjan: I'll update it. I was just ignoring all of the auth checks and owner stuff. We pass the `accessKeyID` around everywhere currently without even using it. I was just going to tackle that all in a F/U

sia/multipart_test.go
```diff
@@ -92,3 +97,82 @@ func TestAbortMultipartUpload(t *testing.T) {
...
+	// prepare a part
+	part := bytes.Repeat([]byte("a"), int(s3.MinUploadPartSize))
+	md5Sum := md5.Sum(part)
+	res2, err := s3Tester.UploadPart(t.Context(), bucket, object, uploadID, 1, part)
+	if err != nil {
+		t.Fatal(err)
+	} else if res2 == nil || res2.ETag == nil || *res2.ETag != s3.FormatETag(md5Sum[:]) {
+		t.Fatalf("unexpected upload part result: %+v", res2)
+	}
+	// TODO: assert various s3 errors for invalid part uploads
```
ChrisSchinnerl, line 134: PR is looking good now but it seems like you wanted to extend this test further

peterjan: Yeah I'm going to F/U with that after I add `ListParts` (and then once more after `CompleteMultipartUpload`)

ChrisSchinnerl (approved):

PR #73, opened 2025-12-05
Add `ListParts`

peterjan:


sia/persist/sqlite/multipart.go
```diff
@@ -108,3 +109,58 @@ func (s *Store) HasMultipartUpload(bucket, name string, uploadID s3.UploadID) er
 		return err
 	})
 }
+
+// ListParts lists uploaded parts for a multipart upload.
+func (s *Store) ListParts(accessKeyID, bucket, name string, uploadID s3.UploadID, partNumberMarker int, maxParts int64) (*s3.ListPartsResult, error) {
+	res := &s3.ListPartsResult{
+		OwnerID:              "", // TODO: sia backend does not yet support owners
+		InitiatorID:          accessKeyID,
```
ChrisSchinnerl: Let's add a todo here as well. This is eventually going to be some unique user id and not the access key id.

sia/multipart.go
```diff
@@ -197,7 +197,18 @@ func (s *Sia) UploadPartCopy(ctx context.Context, accessKeyID, srcBucket, srcObj
 
 // ListParts lists uploaded parts for a multipart upload.
 func (s *Sia) ListParts(ctx context.Context, accessKeyID, bucket, object, uploadID string, page s3.ListPartsPage) (*s3.ListPartsResult, error) {
-	return nil, s3errs.ErrNotImplemented
+	// check bucket access
+	if err := s.store.HeadBucket(accessKeyID, bucket); err != nil {
```
ChrisSchinnerl: We can drop this. `ListParts` is already going to return the right error in case the bucket or upload don't exist. ACL stuff will go into the s3 handler.

ChrisSchinnerl (changes requested):

sia/multipart.go
peterjan: Yeah I know, I'm just being consistently wrong. I figure we follow it up with a PR that removes `accessKeyID` from the interface and all store methods.

ChrisSchinnerl (approved):

PR #76, opened 2025-12-09
Add `ListMultipartUploads`

peterjan:
This PR implements listing for multipart uploads. 

For 100k multipart uploads
```
cpu: Apple M5
BenchmarkListMultipartUploads/no-delim-10                    3148           385382 ns/op
BenchmarkListMultipartUploads/slash-delim-10                 416           2464044 ns/op
BenchmarkListMultipartUploads/random-delim-10                681           2526671 ns/op
BenchmarkListMultipartUploads/paging-10                      328           3211388 ns/op
```

For 1M multipart uploads
```
BenchmarkListMultipartUploads/no-delim-10                     2058          560381 ns/op
BenchmarkListMultipartUploads/slash-delim-10                  60          22588228 ns/op
BenchmarkListMultipartUploads/random-delim-10                 100         19709511 ns/op
BenchmarkListMultipartUploads/paging-10                       100         20316091 ns/op
```

The main difference with objects is that multipart uploads are not unique on a per bucket basis. This has some unfortunate downsides and extra complexity, like upload ID markers alongside key markers. Considering multipart uploads are deleted when they're aborted or completed we should be fine for the time being imo.


sia/persist/sqlite/multipart_test.go
```diff
@@ -114,3 +120,391 @@ func TestAbortMultipartUpload(t *testing.T) {
...
+			}
+		}
+	})
+
+	// benchmark / delimiter with random prefix and paging
+	b.Run("slash-delim-paging", func(b *testing.B) {
+		var cnt int
+		var prevKeyMarker, prevUploadIDMarker string
+		for b.Loop() {
+			prefix := randomPath(1, minLength)
```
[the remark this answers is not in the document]
peterjan: Going to change it slightly but not sure if we care that much.

sia/persist/sqlite/multipart.go
```diff
@@ -116,3 +119,220 @@ func (s *Store) HasMultipartUpload(bucket, name, uploadID string) error {
...
+		searchOffset = prefixLen + 1
+	)
+
+	// check bucket
+	where := []string{"bucket_id = ?"}
+	args := []any{bucketID}
+
+	// handle prefix
+	if hasPrefix {
+		where = append(where, "SUBSTR(name, 1, ?) = ?")
```
chris124567: I think you can make this and the other prefix searches faster with:

```
                where = append(where, ` AND name >= ? AND name < ?`)
                args = append(args, prefix, prefix+"\xFF")
```

sia/persist/sqlite/migrations_test.go
```diff
@@ -45,6 +45,8 @@ CREATE TABLE multipart_uploads (
     FOREIGN KEY (bucket_id) REFERENCES buckets(id)
 );
 CREATE INDEX multipart_uploads_bucket_id_name_idx ON multipart_uploads(bucket_id, name);
+CREATE INDEX multipart_uploads_list_idx ON multipart_uploads(bucket_id, name, upload_id, created_at);
+CREATE INDEX multipart_uploads_slash_delim_idx ON multipart_uploads(bucket_id, name, upload_id, created_at) WHERE INSTR(name, '/') > 0;
```
ChrisSchinnerl: Have you compared how Christopher's approach compares when applied to multipart uploads? The benchmarks seem worse but that alone is not a fair comparison. 

An index over all rows containing a `/` seems overly optimized for a very specific use case too.

sia/persist/sqlite/multipart.go
peterjan: Yeah it's a lot faster. Good job. I was a little apprehensive because I'm not 100% positive that trick stands up to UTF8 prefixes 😢 . Maybe it does though, tests are passing and it's 3x faster so I say we should go for it.

sia/persist/sqlite/migrations_test.go
peterjan: Yeah we can't compare benchmarks between list objects and multipart unfortunately. The two are really very different and I think listing multipart uploads will always be slower. On the bright side though we need more scale from objects than we do multipart uploads.

Correct me if I'm wrong @chris124567 , but if I read through the comments on your PR I think adding the walk test surfaced a bug in the alternative approach right? Where a lot of the nodes weren't visited (exactly once). I believe you mention the duplicate key thing but that was only for the cases where you were missing 3 or 4 right? 

I removed the index.

peterjan: Benchmarks improved 3x by the way, by applying the 0xFF trick to all prefix checks.

chris124567: > Yeah we can't compare benchmarks between list objects and multipart unfortunately. The two are really very different and I think listing multipart uploads will always be slower. On the bright side though we need more scale from objects than we do multipart uploads.
> 
> 
> 
> Correct me if I'm wrong @chris124567 , but if I read through the comments on your PR I think adding the walk test surfaced a bug in the alternative approach right? Where a lot of the nodes weren't visited (exactly once). I believe you mention the duplicate key thing but that was only for the cases where you were missing 3 or 4 right? 
> 
> 
> 
> I removed the index. 
> 
> 

Yeah @peterjan there were bugs found by your test.  Some keys would be missed every time. I spent much of this past weekend trying to get it to pass unsuccessfully.

peterjan: In that case I'm not sure if it's worth exploring.

sia/persist/sqlite/multipart.go
```diff
@@ -65,17 +71,24 @@ func (s *Store) AddMultipartPart(bucket, name string, uploadID s3.UploadID, file
...
 		var sha256Value any
 		if contentSHA256 != nil {
 			sha256Value = sqlHash256(*contentSHA256)
 		}
 
-		err = tx.QueryRow(`SELECT filename FROM multipart_parts WHERE multipart_upload_id = $1 AND part_number = $2`, uid, partNumber).Scan(&prevFilename)
+		err = tx.QueryRow(`SELECT filename FROM multipart_parts WHERE multipart_upload_id = $1 AND part_number = $2`,
+			sqlUploadID(uploadID), partNumber).Scan(&prevFilename)
 		if err != nil && !errors.Is(err, sql.ErrNoRows) {
 			return err
```
ChrisSchinnerl, line 93: No special error for missing parts?

sia/persist/sqlite/multipart.go
```diff
@@ -157,10 +183,208 @@ func (s *Store) ListParts(bucket, name string, uploadID s3.UploadID, partNumberM
...
+	// set default max uploads
+	maxUploads := page.MaxUploads
+	if maxUploads == 0 {
+		maxUploads = 1000
+	}
+
+	if prefix.HasDelimiter {
+		return s.listMultipartUploadsWithDelim(bucket, prefix, keyMarker, uploadIDMarker, maxUploads)
+	}
+	return s.listMultipartUploadsNoDelim(bucket, prefix.Prefix, keyMarker, uploadIDMarker, maxUploads)
```
ChrisSchinnerl: Unless there is a huge benefit to it I'd avoid splitting the logic here. It seems like `listMultipartUploadsWithDelim` can handle both cases because `Prefix` itself can. 

If we know that there is no delimiter we can even change the batch limit of `100` to `maxUploads` to make it run in a single iteration. Knowing that we will not end up wasting time going down a directory.

sia/persist/sqlite/init.sql
```diff
@@ -23,26 +23,26 @@ CREATE TABLE objects (
...
-    upload_id BLOB NOT NULL UNIQUE,
+    upload_id BLOB PRIMARY KEY,
     bucket_id INTEGER NOT NULL,
     name TEXT NOT NULL,
     metadata TEXT NOT NULL,
     created_at INTEGER NOT NULL,
     FOREIGN KEY (bucket_id) REFERENCES buckets(id)
 );
 CREATE INDEX multipart_uploads_bucket_id_name_idx ON multipart_uploads(bucket_id, name);
+CREATE INDEX multipart_uploads_bucket_id_name_upload_id_created_at_idx ON multipart_uploads(bucket_id, name, upload_id, created_at);
```
ChrisSchinnerl: You should be able to skip `created_at` here since you don't order by it anyway. It's only selected from what I can tell.

ChrisSchinnerl (changes requested):

peterjan:
This is ListObjects on an M5

| Benchmark                                        | Time (ms/op) |
|-------------------------------------------------|--------------|
| BenchmarkListObjects/no_delimiter_no_prefix      | 0.75         |
| BenchmarkListObjects/root_delimiter              | 9.00         |
| BenchmarkListObjects/random_without_delimiter    | 0.09         |
| BenchmarkListObjects/random_with_root_delimiter  | 0.13         |
| BenchmarkListObjects/folder_bottom_delimiter     | 0.03         |
| BenchmarkListObjects/folder_delimiter            | 0.13         |

This is ListMultipartObjects on an M5
| Benchmark                                                     | Time (ms/op) |
|--------------------------------------------------------------|--------------|
| BenchmarkListMultipartUploads/no-delim                        | 0.49         |
| BenchmarkListMultipartUploads/slash-delim-no-prefix           | 23.68        |
| BenchmarkListMultipartUploads/slash-delim-rand-prefix         | 9.08         |
| BenchmarkListMultipartUploads/slash-delim-paging              | 38.64        |
| BenchmarkListMultipartUploads/random-delim                    | 1.34         |

sia/persist/sqlite/multipart.go
peterjan: There doesn't have to be an existing part right. This query returns the previous filename so we can delete it after we overwrote the part in the db

sia/persist/sqlite/multipart.go
peterjan: That's what I had initially and I preferred the split. I can change it back though

chris124567 (approved):

ChrisSchinnerl (approved):

PR #77, opened 2025-12-11
Add `UploadPartCopy`

peterjan:


sia/multipart.go
```diff
@@ -192,7 +192,119 @@ func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object, uploa
...
+	} else if err != nil {
+		return nil, fmt.Errorf("failed to open part directory: %w", err)
+	} else if err := errors.Join(dir.Sync(), dir.Close()); err != nil {
+		return nil, fmt.Errorf("failed to sync part directory: %w", err)
+	}
+
+	// add multipart part to the database
+	previous, err := s.store.AddMultipartPart(dstBucket, dstObject, uid, filepath.Base(partPath), opts.PartNumber, contentMD5, nil, contentLength)
+	if err != nil {
+		if err := os.Remove(partPath); err != nil {
```
ChrisSchinnerl: I think this is redundant since you defer it already. Not sure if we missed that on a previous PR as well.

ChrisSchinnerl (changes requested):

sia/multipart.go
peterjan: right you are

ChrisSchinnerl (approved):

PR #78, opened 2025-12-12
Add `CompleteMultipartUpload`

### LINKED ISSUE #48: Implement multipart uploads in Sia backend

peterjan:
Closes https://github.com/SiaFoundation/s3d/issues/48


sia/persist/sqlite/init.sql
```diff
@@ -33,18 +33,34 @@ CREATE TABLE multipart_uploads (
...
+
+    -- one of these foreign keys must be set
+    multipart_upload_id INTEGER,
+    object_id INTEGER,
+
+    -- enforce atomicity
+    CHECK (
+        (multipart_upload_id IS NOT NULL AND object_id IS NULL AND created_at IS NOT NULL AND filename IS NOT NULL AND offset IS NULL) OR
+        (multipart_upload_id IS NULL AND object_id IS NOT NULL AND created_at IS NULL AND filename IS NULL AND offset IS NOT NULL)
+    ),
```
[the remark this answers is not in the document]
peterjan: outdated

internal/testutil/backend.go
```diff
@@ -851,3 +870,21 @@ func (o object) matches(oid s3.ObjectID) bool {
...
+	if s == "" {
+		return [16]byte{}
+	}
+
+	var etag [16]byte
+	decoded, err := hex.DecodeString(s)
+	if err != nil {
+		return [16]byte{}
+	}
+	copy(etag[:], decoded)
```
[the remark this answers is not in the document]
peterjan: That's fine, this will error out with `ErrInvalidpart` which is what we want.

sia/multipart.go
```diff
@@ -207,6 +210,106 @@ func (s *Sia) ListParts(ctx context.Context, accessKeyID, bucket, object, upload
...
+		lookup[part.PartNumber] = part
+	}
+
+	// deduplicate parts
+	for i := 1; i < len(parts); i++ {
+		if parts[i].PartNumber == parts[i-1].PartNumber {
+			parts = append(parts[:i], parts[i+1:]...)
+			i--
+		}
+	}
```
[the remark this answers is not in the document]
peterjan: you'd be left with one part, I added a NOTE saying parts must be between 1 and max parts, this is already validated in the handler

internal/testutil/backend.go
```diff
@@ -680,41 +682,54 @@ func (b *MemoryBackend) CompleteMultipartUpload(_ context.Context, accessKeyID,
 		return nil, s3errs.ErrNoSuchUpload
 	}
 
-	if len(parts) == 0 {
-		return nil, s3errs.ErrInvalidRequest
+	// deduplicate parts, keeping the last occurrence
```
ChrisSchinnerl: This is duplicate logic between the memory and Sia implementation. Same as the call to `UploadIDFromString`. If the id needs to be parsed anyway you can do so in the handler already. Same goes for compacting the parts. Then the backend doesn't need to worry about it.

s3/s3.go
```diff
@@ -223,9 +223,11 @@ type Backend interface {
 	// - If any referenced part is missing or its ETag does not match,
 	//   [ErrInvalidPart] must be returned.
 	//
-	// NOTE: the parts slice is expected to be sorted by part number and contain
-	// no duplicates.
-	CompleteMultipartUpload(ctx context.Context, accessKeyID, bucket, object, uploadID string, parts []CompletedPart) (*CompleteMultipartUploadResult, error)
+	// - If the part numbers of the parts are not provided in ascending order,
+	//   [ErrInvalidPartOrder] must be returned.
+	//
+	// - If a part is too small, [ErrEntityTooSmall] must be returned.
```
ChrisSchinnerl: Might want to clarify here that all parts but the last one need to have the min size.

s3/s3.go
```diff
@@ -223,9 +223,11 @@ type Backend interface {
 	// - If any referenced part is missing or its ETag does not match,
 	//   [ErrInvalidPart] must be returned.
 	//
-	// NOTE: the parts slice is expected to be sorted by part number and contain
-	// no duplicates.
-	CompleteMultipartUpload(ctx context.Context, accessKeyID, bucket, object, uploadID string, parts []CompletedPart) (*CompleteMultipartUploadResult, error)
+	// - If the part numbers of the parts are not provided in ascending order,
```
ChrisSchinnerl, line 226: Can be moved to the handler and ignored here.

sia/objects/reader.go
ChrisSchinnerl, line 1: `Reader` and `reader.go` are a bit generic for what this is. `MultipartReader` might be better.

sia/objects/reader.go
```diff
@@ -0,0 +1,179 @@
...
+// Reader provides a sequential reader over the parts of a multipart upload, it
+// is not safe for concurrent use.
+type Reader struct {
+	parts    []Part
+	basePath string
+
+	curr      *os.File
+	currIdx   int
+	bytesLeft int64
+	totalSize int64
```
ChrisSchinnerl: `totalSize` is unnecessary since the only time you call `Size` on the reader you already know the total size.

sia/objects/reader.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		n, err := r.curr.Read(p[:toRead])
+		if n > 0 {
+			r.bytesLeft -= int64(n)
+			if r.partHash != nil {
+				_, _ = r.partHash.Write(p[:n])
+			}
+			if r.totalHash != nil {
+				_, _ = r.totalHash.Write(p[:n])
```
ChrisSchinnerl: I don't think that's how S3 computes the final hash. [These docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/tutorial-s3-mpu-additional-checksums.html#verify-object-integrity-step8) mention that the final etag is not the md5 hash of the data but a composite of the individual part etags.

In which case the caller can already compute the final hash and the reader only has to validate the individual part etags.

sia/objects/reader.go
```diff
@@ -0,0 +1,179 @@
...
+	"path/filepath"
+	"strconv"
+
+	"github.com/SiaFoundation/s3d/s3/s3errs"
+)
+
+// Reader provides a sequential reader over the parts of a multipart upload, it
+// is not safe for concurrent use.
+type Reader struct {
+	parts    []Part
```
ChrisSchinnerl: If you turn this into `remainingParts` and just pop off its front whenever you call `finishPart` you don't need `currIdx`.

sia/objects/reader.go
```diff
@@ -0,0 +1,179 @@
...
+	basePath string
+
+	curr      *os.File
+	currIdx   int
+	bytesLeft int64
```
ChrisSchinnerl: I'd drop this too. You already check the size of each part directly after opening it. So you can rely on `io.EOF` to figure out when the part is finished and since you check the hash afterwards you don't need to be afraid of corruption, too few or too many bytes.

sia/objects/reader.go
```diff
@@ -0,0 +1,179 @@
...
+		parts:     parts,
+		basePath:  basePath,
+		currIdx:   -1,
+		totalSize: total,
+		totalHash: md5.New(),
+	}, nil
+}
+
+// Read reads data from the multipart upload parts sequentially.
+func (r *Reader) Read(p []byte) (int, error) {
```
ChrisSchinnerl: This is quite a bit more complicated than it needs to be I think. You should be able to get away with something like the following:

```go
func (r *Reader) Read(p []byte) (int, error) {
	// no current part, open next
	if r.curr == nil {
		if err := r.openNext(); err != nil {
			return 0, err // io.EOF if no more parts
		}
	}

	n, err := r.curr.Read(p)
	if n > 0 {
		_, _ = r.partHash.Write(p[:n])
	}
	if errors.Is(err, io.EOF) {
		return n, r.finishPart() // ignore EOF, try next part
	}
	return n, err
}
```

This effectively drops `bytesLeft` in favor of checking for `io.EOF` and drops the complexity of the loop. Any caller will call `Read` in a loop anyway until `io.EOF`.

sia/objects/reader.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	r.currIdx++
+	if r.currIdx >= len(r.parts) {
+		return io.EOF
+	}
+
+	p := r.parts[r.currIdx]
+	path := filepath.Join(r.basePath, strconv.Itoa(p.PartNumber), p.Filename)
+	f, err := os.Open(path)
```
ChrisSchinnerl: If you change this to `r.curr, err = os.Open(path)` you don't need to call `f.Close` on every error since the caller will call `Close` on the reader anyway.

sia/objects/reader.go
```diff
@@ -0,0 +1,179 @@
...
+		return n, err
+	}
+}
+
+// Close closes the reader and any open part file.
+func (r *Reader) Close() error {
+	if r.curr != nil {
+		return r.curr.Close()
+	}
+	return nil
```
ChrisSchinnerl: You can do `return r.curr.Close` since `Close` on a `os.File` will return an error if the file handle is `nil` instead of panicking.

sia/objects/reader.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func (r *Reader) finishPart() error {
+	if r.partHash == nil {
+		return nil
+	}
+	var sum [16]byte
+	copy(sum[:], r.partHash.Sum(nil))
+	if sum != r.parts[r.currIdx].ContentMD5 {
```
ChrisSchinnerl: (nit) you can save a copy here by comparing slices directly 

```if sum := r.partHash.Sum(nil); !bytes.Equal(sum, r.parts[r.currIdx].ContentMD5[:]) {```

File: sia/persist/sqlite/init.sql [hunk 15]
```diff
@@ -18,11 +18,12 @@ CREATE TABLE objects (
...
     size INTEGER NOT NULL,
     updated_at INTEGER NOT NULL,
-    PRIMARY KEY(bucket_id, name)
+    PRIMARY KEY (bucket_id, name)
 ) WITHOUT ROWID;
 
 CREATE TABLE multipart_uploads (
-    upload_id BLOB PRIMARY KEY,
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    upload_id BLOB NOT NULL UNIQUE,
```
ChrisSchinnerl: Is there a reason for this change?

sia/persist/sqlite/init.sql
```diff
@@ -32,18 +33,35 @@ CREATE TABLE multipart_uploads (
...
     content_length INTEGER NOT NULL,
-    created_at INTEGER NOT NULL,
-    FOREIGN KEY (multipart_upload_id) REFERENCES multipart_uploads(upload_id) ON DELETE CASCADE,
-    UNIQUE(multipart_upload_id, part_number)
+
+    created_at INTEGER, -- nulled when the part is finalized
+    filename TEXT, -- nulled when the part is finalized
+    offset INTEGER, -- set when the part is finalized
+
+    -- one of these foreign keys must be set
```
ChrisSchinnerl: This is quite nasty. Especially the checks we need to enforce it and the fact we need 2 columns to reference an object.

I'd honestly rather go for a separate `object_parts` table which references `objects`. Then we cleanly separate parts of finished uploads from the ones that are ongoing or even stale.

sia/persist/sqlite/multipart.go
```diff
@@ -25,21 +26,122 @@ func (s *Store) CreateMultipartUpload(bucket, name string, uploadID s3.UploadID,
...
+
+		// get multipart upload id
+		var uid int64
+		err = tx.QueryRow(`SELECT id FROM multipart_uploads WHERE upload_id = $1 AND bucket_id = $2 AND name = $3`,
+			sqlUploadID(uploadID), bid, name).Scan(&uid)
+		if err != nil {
+			return err
+		}
+
+		// validate parts exist and are continuous
```
ChrisSchinnerl: (nit)
```suggestion
		// validate parts exist and are contiguous
```

sia/persist/sqlite/multipart_test.go
```diff
@@ -243,6 +243,86 @@ func TestListParts(t *testing.T) {
...
+	}
+
+	partMD5A := frand.Entropy128()
+	partMD5B := frand.Entropy128()
+	parts := []objects.Part{
+		{PartNumber: 1, Filename: "part-1", Size: s3.MinUploadPartSize, ContentMD5: partMD5A},
+		{PartNumber: 2, Filename: "part-2", Size: 5, ContentMD5: partMD5B}, // Last part can be any size
+	}
+
+	// Add parts to the upload
```
ChrisSchinnerl: Might want to tell Claude to be consistent with upper vs lowercase in comments.

sia/persist/sqlite/objects.go
```diff
@@ -51,51 +51,65 @@ func (s *Store) DeleteObject(accessKeyID, bucket string, objectID s3.ObjectID) e
...
+				SELECT object_id, metadata, updated_at, size, content_md5
+				FROM objects
+				WHERE bucket_id = $1 AND name = $2
+			`, bid, name).Scan((*sqlHash256)(&obj.ID), (*sqlMetaJSON)(&obj.Meta), (*sqlTime)(&obj.LastModified), &obj.Length, (*sqlMD5)(&obj.ContentMD5))
 		}
-		return nil
-	})
-	return &obj, err
+
+		// part specified - validate and return single part
```
ChrisSchinnerl: What validation?

sia/multipart.go
```diff
@@ -321,8 +322,111 @@ func (s *Sia) ListParts(ctx context.Context, accessKeyID, bucket, object, upload
 }
 
 // CompleteMultipartUpload completes a multipart upload.
-func (s *Sia) CompleteMultipartUpload(ctx context.Context, accessKeyID, bucket, object, uploadID string, parts []s3.CompletedPart) (*s3.CompleteMultipartUploadResult, error) {
-	return nil, s3errs.ErrNotImplemented
+//
+// NOTE: parts must be non-nil and not exceed MaxParts.
```
ChrisSchinnerl: The working is a bit unclear. Do you mean the parts have to be checked for those requirements? In that case it's redundant here and in the memory backend since the handler guarantees this. If you meant to say that it's safe to assume this I'd word it slightly differently.

sia/multipart_test.go
```diff
@@ -279,6 +279,75 @@ func TestMultipartListParts(t *testing.T) {
 	}
 }
 
+func TestCompleteMultipartUpload(t *testing.T) {
```
ChrisSchinnerl, line 282: This is missing a few of the more obvious edge cases like small parts, parts being out of order or invalid etags.

ChrisSchinnerl (changes requested):

sia/multipart_test.go
peterjan: I had a comment in the PR description saying I'd F/U with more testing. The PR was getting a bit out of hand LoC wise. I split off the MultipartReader, depending on the size of the PR I'll either add it here or F/U with extended tests to cover all of the edge cases.

sia/multipart.go
peterjan: Resolving this in favour of https://github.com/SiaFoundation/s3d/pull/78/changes#r2664825717 - I'll reply there since it's very much related.

s3/s3.go
peterjan: Resolving in favour of https://github.com/SiaFoundation/s3d/pull/78/changes#r2664825717

internal/testutil/backend.go
peterjan: I want to keep the `UploadIDFromString` for now. You are right though I already considered that but I would like to change all occurrences at the same time, e.g. if we change it here the entire backend interface should require an uploadID instead of a string.

peterjan: I had a good reason to do the parts validation in the backends. I don't remember it clearly though, it's been quite some time. I think the reason is twofold: a) the test suite is very particular in what error it expects, e.g. we can't return `s3errs.ErrInvalidPart` when it expects `s3errs.ErrInvalidPartOrder` and b) in the Sia backend we have to fetch the parts from the store to verify the given parts on complete against the ones we were given during the upload. I remember moving this around quite a few times because I was also a little annoyed to have to move it into the backend. I'll check it out.

edit: you only mentioned compacting the parts - I see that now 😂

sia/persist/sqlite/init.sql
peterjan: I think this is a leftover from playing around with `WITHOUT ROWID` when I was benchmarking `ListMultipartUploads`. I got rid of it, thanks for catching that. I made sure to re-run the benchmark to catch any possible regression and there is none in the latest version.

sia/persist/sqlite/init.sql
peterjan: Here I am thinking I'm all fancy

s3/multipart.go
```diff
@@ -154,16 +154,15 @@ func NewUploadID() (uid UploadID) {
 	return
 }
 
-// UploadIDFromString parses an upload ID from its hexadecimal string
-// representation.
-func UploadIDFromString(s string) (UploadID, error) {
-	var uid UploadID
+// ParseUploadID parses an upload ID from its hexadecimal string representation.
+// If the string is not a valid upload ID, an empty UploadID is returned.
+func ParseUploadID(s string) (uid UploadID) {
```
ChrisSchinnerl: Should return an error. You can call this in `routeMultipartUpload` and handle the error there in a single spot.

ChrisSchinnerl (changes requested):
Got a bunch of conflicts now

s3/multipart.go
peterjan: I don't think it should. The s3-tests pass junk as an upload ID and expects `ErrNoSuchUpload` instead of a 400 invalid argument.

ChrisSchinnerl: Then return `ErrNoSuchUpload` if you can't parse it?

ChrisSchinnerl (approved):

PR #83, opened 2025-12-22
Add `ListObjects` to Sia backend

peterjan:


chris124567 (approved):

sia/objects.go
```diff
@@ -160,7 +160,17 @@ func (s *Sia) headOrGetObject(ctx context.Context, accessKeyID *string, bucket,
 // contents of the bucket and sort the results into the Contents and
 // CommonPrefixes fields of the returned ObjectsListResult.
 func (s *Sia) ListObjects(ctx context.Context, accessKeyID *string, bucket string, prefix s3.Prefix, page s3.ListObjectsPage) (*s3.ObjectsListResult, error) {
-	return nil, s3errs.ErrNotImplemented
+	if accessKeyID == nil {
+		// anonymous access is not supported yet
+		return nil, s3errs.ErrAccessDenied
+	}
+
+	// quick check if the bucket exists
```
chris124567, line 168: nit: I think this is not necessary because we already error in this case

https://github.com/SiaFoundation/s3d/blob/master/sia/persist/sqlite/objects.go#L138

https://github.com/SiaFoundation/s3d/blob/master/internal/testutil/backend.go#L253

peterjan: Hm yeah that's copy pasta in a way, the comment is a little unfortunate. The `HeadBucket` method also asserts the access key has access to the bucket (in the memory backend that is, Sia backend doesn't have that yet). Instead of changing it here I'd rather F/U with a PR that tackles all these ACL checks. All of our store methods for instance take an access key ID but don't use it. Don't mind changing it though if you all feel differently, definitely we have to F/U with an update to our auth code at some point soon.

chris124567: Follow up is fine by me :+1:

ChrisSchinnerl (approved):

PR #87, opened 2026-01-06
Add SDK opts

peterjan:
Playing around with s3d I needed to get around the default upload/download options. Especially the default host timeout on upload of 4s is way too strict imo.

### DIFF
--- a/sia/sdk.go
+++ b/sia/sdk.go
@@ -9,35 +9,55 @@ import (
 	"go.sia.tech/indexd/sdk"
 )
 
-// IndexdSDK is a wrapper around the indexd SDK to implement the SDK interface.
-type IndexdSDK struct {
-	inner *sdk.SDK
+type (
+	// IndexdSDK is a wrapper around the indexd SDK to implement the SDK interface.
+	IndexdSDK struct {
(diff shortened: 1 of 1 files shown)


sia/sdk.go
```diff
@@ -9,35 +9,55 @@ import (
...
+		opt(indexd)
+	}
+	return indexd
 }
 
 // Download downloads an object from indexd.
 func (s *IndexdSDK) Download(ctx context.Context, w io.Writer, obj sdk.Object, rnge *s3.ObjectRange) error {
-	var opts []sdk.DownloadOption
 	if rnge != nil {
-		opts = append(opts, sdk.WithDownloadRange(uint64(rnge.Start), uint64(rnge.Length)))
```
ChrisSchinnerl, line 31: I'd go the safe route here and deep clone to avoid the nasty edge case of `s.dlOpts` potentially having a capacity > length and `append` reusing the underlying memory. Which might be alright but strictly speaking is a race.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #89, opened 2026-01-06
Allow empty objects

peterjan:
In S3 it's perfectly legal to upload an empty object, our indexer doesn't allow saving objects without slabs so we should handle it in s3d and skip the `sdk.Upload`. We're storing empty Object IDs though, which is not ideal but not a big deal I guess?

### DIFF
--- a/sia/objects.go
+++ b/sia/objects.go
@@ -15,6 +15,7 @@ import (
 	"github.com/SiaFoundation/s3d/s3"
 	"github.com/SiaFoundation/s3d/s3/s3errs"
 	"github.com/SiaFoundation/s3d/sia/objects"
+	"go.sia.tech/core/types"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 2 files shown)


sia/objects.go
```diff
@@ -196,10 +203,20 @@ func (s *Sia) PutObject(ctx context.Context, accessKeyID string, bucket, object
...
 
-	// upload the data
-	obj, err := s.sdk.Upload(ctx, lr)
-	if err != nil {
-		return nil, fmt.Errorf("failed to upload object: %w", err)
+	// handle empty object case
+	var objectID types.Hash256
+	if opts.ContentLength == 0 {
+		// drain reader
+		if _, err := io.Copy(io.Discard, lr); err != nil {
```
ChrisSchinnerl, line 210: We can save a few lines here if we just don't drain the reader. There is no point in doing so when the body is empty. Draining would only be relevant if it was a non-empty http body.

ChrisSchinnerl (changes requested):

sia/objects.go
peterjan: I think we should drain the reader because it doesn't matter if the body is empty, but should it not be empty we want it to fail the `opts.ContentMD5` assertion on line 234? I just wanted to be correct here, I agree it's a bit of a silly edge case but what's the harm in draining the reader here.

ChrisSchinnerl: I'm quite sure that's not possible. Because afaik the Go http server won't allow you to read more bytes from the body if content length is known. Which we guarantee. But I don't mind keeping it to be safe.

ChrisSchinnerl (approved):

PR #93, opened 2026-01-07
Add Multipart reader

peterjan:
Extracted from https://github.com/SiaFoundation/s3d/pull/78

### DIFF
--- a/sia/objects/multipartreader.go
+++ b/sia/objects/multipartreader.go
@@ -59,6 +59,9 @@ func (r *MultipartReader) Read(p []byte) (int, error) {
 
 // Close closes the reader and any open part file.
 func (r *MultipartReader) Close() error {
+	if r.curr == nil {
+		return nil
+	}
 	return r.curr.Close()
 }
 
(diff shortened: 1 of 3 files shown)


sia/objects/multipartreader.go
```diff
@@ -0,0 +1,117 @@
...
+		return fmt.Errorf("failed to open part at %s: %w", path, err)
+	}
+
+	stat, err := r.curr.Stat()
+	if err != nil {
+		r.curr.Close()
+		return fmt.Errorf("failed to stat part %d: %w", p.PartNumber, err)
+	}
+	if stat.Size() != p.Size {
+		r.curr.Close()
```
ChrisSchinnerl, line 90: Strictly speaking not necessary. Copilot probably just lacks context of the overall use of the `MultipartReader`. Doesn't hurt either though.

sia/objects/multipartreader.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (r *MultipartReader) finishPart() error {
+	if err := r.curr.Close(); err != nil {
+		return fmt.Errorf("failed to close part file: %w", err)
+	}
+	r.curr = nil
+
+	if len(r.remainingParts) == 0 {
+		return io.EOF
```
ChrisSchinnerl: Can this ever happen? When you were able to open a part then you should have a part in here as well.
So I'd return a dedicated error and not `io.EOF`.

sia/objects/multipartreader.go
peterjan: Might as well keep it if AI is going to keep complaining about it, doesn't really hurt.

sia/objects/multipartreader.go
peterjan: Hm what about instead we keep a `currPart` on the reader. That way we don't split the dequeueing across two methods (open/finish) and we avoid the dedicated error that should never happen. Did that for the time being, feel free to reopen.

ChrisSchinnerl (approved):

PR #95, opened 2026-01-22
Use cached slab metadata for downloads rather than fetching it from the indexer

### LINKED ISSUE #90: Use cached slab metadata for downloads rather than fetching it from the indexer
`headOrGetObject` currently has a TODO which we can resolve now that the full metadata is returned for objects. 

What that means is instead of calling `s.sdk.Object` on every download we persist a timestamp alongside the object in the database to indicate how old the metadata is. If it is older than some constant (24 hours is probably a reasonable start) we refresh it on demand. If refreshing it fails, we try the outdated metadata instead.

chris124567:
Close #90


sia/persist/sqlite/objects.go
```diff
@@ -61,12 +61,12 @@ func (s *Store) GetObject(accessKeyID *string, bucket, name string) (*objects.Ob
...
 		var meta string
 		err = tx.QueryRow(`
-			SELECT name, object_id, content_md5, metadata, size, updated_at
+			SELECT name, object_id, content_md5, metadata, size, updated_at, object, object_retrieved
 			FROM objects
 			WHERE bucket_id = $1 AND name = $2
 		`, bid, name).
 			Scan(&obj.Name, (*sqlHash256)(&obj.ID), (*sqlMD5)(&obj.ContentMD5), &meta,
-				&obj.Size, (*sqlTime)(&obj.UpdatedAt))
+				&obj.Size, (*sqlTime)(&obj.UpdatedAt), &obj.Object, (*sqlTime)(&obj.ObjectRetrieved))
```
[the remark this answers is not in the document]
chris124567: we aren't doing migrations yet

sia/persist/sqlite/init.sql
```diff
@@ -18,6 +18,8 @@ CREATE TABLE objects (
     metadata TEXT NOT NULL,
     size INTEGER NOT NULL,
     updated_at INTEGER NOT NULL,
+    object TEXT NOT NULL,
+    object_retrieved INTEGER NOT NULL,
```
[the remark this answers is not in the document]
chris124567: we aren't doing migrations yet

sia/persist/sqlite/objects.go
```diff
@@ -99,16 +99,18 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj *objects.Object)
...
 				content_md5 = excluded.content_md5,
 				metadata = excluded.metadata,
 				size = excluded.size,
-				updated_at = excluded.updated_at
+				updated_at = excluded.updated_at,
+				object = excluded.object,
+				object_retrieved = excluded.object_retrieved
 		`, bid, name, sqlHash256(obj.ID), sqlMD5(obj.ContentMD5),
-			string(metaJson), obj.Size, sqlTime(obj.UpdatedAt))
+			string(metaJson), obj.Size, sqlTime(obj.UpdatedAt), string(obj.Object), sqlTime(obj.ObjectRetrieved))
```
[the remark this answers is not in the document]
chris124567: we aren't doing migrations yet

sia/persist/sqlite/migrations.go
```diff
@@ -1,6 +1,8 @@
 package sqlite
 
-import "go.uber.org/zap"
+import (
+	"go.uber.org/zap"
+)
 
 // migrations is a list of functions that are run to migrate the database from
```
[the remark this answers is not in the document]
chris124567: we aren't doing migrations yet

sia/objects.go
```diff
@@ -139,11 +143,37 @@ func (s *Sia) headOrGetObject(ctx context.Context, accessKeyID *string, bucket,
...
+	// fetch from indexer if cache is missing or expired
+	if !cached || time.Since(obj.ObjectRetrieved) >= metadataCacheLifetime {
+		fetched, err := s.sdk.Object(ctx, obj.ID)
+		if err != nil {
+			if cached {
+				s.logger.Warn("failed to fetch object from indexer, using stale metadata", zap.Error(err))
+			} else {
+				return nil, fmt.Errorf("failed to fetch object from indexer: %w", err)
+			}
+		} else {
```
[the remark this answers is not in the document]
chris124567: added tests

sia/persist/sqlite/migrations.go
```diff
@@ -1,6 +1,8 @@
 package sqlite
 
-import "go.uber.org/zap"
+import (
+	"go.uber.org/zap"
+)
```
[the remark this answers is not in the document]
chris124567: we aren't doing migrations yet

sia/persist/sqlite/migrations_test.go
```diff
@@ -31,6 +31,8 @@ CREATE TABLE objects (
     metadata TEXT NOT NULL,
     size INTEGER NOT NULL,
     updated_at INTEGER NOT NULL,
+    object TEXT NOT NULL,
+    object_retrieved INTEGER NOT NULL,
```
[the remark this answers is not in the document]
chris124567: we aren't doing migrations yet

sia/objects.go
```diff
@@ -244,6 +293,9 @@ func (s *Sia) PutObject(ctx context.Context, accessKeyID string, bucket, object
 	}); err != nil {
 		return nil, fmt.Errorf("failed to store object metadata: %w", err)
 	}
+	if err := s.store.UpdateObjectCache(accessKeyID, bucket, object, encoded, time.Now()); err != nil {
+		return nil, fmt.Errorf("failed to update object cache: %w", err)
+	}
```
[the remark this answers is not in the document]
chris124567: I disagree because we don't really take this approach in the rest of PutObject or most of the other member functions even though we do in getCachedObject

sia/persist/sqlite/objects.go
```diff
@@ -50,6 +51,28 @@ func (s *Store) DeleteObject(accessKeyID, bucket string, objectID s3.ObjectID) e
 	})
 }
 
+// GetCachedObject retrieves only the cache-related fields for an object.
+func (s *Store) GetCachedObject(accessKeyID, bucket, name string) (objectID types.Hash256, object []byte, objectRetrieved time.Time, err error) {
```
peterjan: Probably cleaner to add a type for `sdk.Object` in our types file.

sia/persist/sqlite/objects.go
```diff
@@ -99,20 +122,45 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj *objects.Object)
 		}
 
 		_, err = tx.Exec(`
-			INSERT INTO objects (bucket_id, name, object_id, content_md5, metadata, size, updated_at)
-			VALUES ($1, $2, $3, $4, $5, $6, $7)
+			INSERT INTO objects (bucket_id, name, object_id, content_md5, metadata, size, updated_at, object, object_retrieved)
+			VALUES ($1, $2, $3, $4, $5, $6, $7, "", 0)
 			ON CONFLICT(bucket_id, name) DO UPDATE SET
```
peterjan, line 127: our `UPDATE` ignores the cache fields here, we should unset the cache fields

sia/persist/sqlite/init.sql
peterjan: I'm wondering whether this should have been a separate table... I personally feel that is cleaner, but this is the type of thing that we can easily disagree on. The main reason I think we should have a separate table:

- in `UpdateCachedObject` we have a note in the docstring indicating we only update the cache fields, but leave the `updated_at` field untouched, that's a bit of a codesmell imo
- I'm not a big fan of the column names.. object and object_retrieved, if we have a separate table `object_retrieved` can be `created_at` and `updated_at` which hold more context since we have two timestamps and the names become self explanatory
- we can clear cache by pruning a table, always nice
- when we update the object we can delete the corresponding row to clear from cache

sia/objects.go
```diff
@@ -244,6 +293,9 @@ func (s *Sia) PutObject(ctx context.Context, accessKeyID string, bucket, object
 	}); err != nil {
 		return nil, fmt.Errorf("failed to store object metadata: %w", err)
 	}
+	if err := s.store.UpdateCachedObject(accessKeyID, bucket, object, encoded, time.Now()); err != nil {
```
peterjan: It's a little odd that this is two separate calls to the store. One to add the object and then one to cache its slab metadata. If you look at the callers for `PutObject` you'll see that `CopyObject` does it too, meaning we missed the `UpdateCachedObject` call there. 

What's preventing us from nesting `sdk.Object` in our object? Don't we always have this information? Every object we upload passes through the SDK right? With this in mind it might make more sense to keep this information on the `objects` table and _not_ add a separate table for the cached fields..

sia/persist/sqlite/migrations_test.go
peterjan: These names are confusing imo. `cached_metadata` and `cached_at` perhaps?

sia/persist/sqlite/objects.go
```diff
@@ -99,20 +122,45 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj *objects.Object)
...
 				updated_at = excluded.updated_at
-		`, bid, name, sqlHash256(obj.ID), sqlMD5(obj.ContentMD5),
-			string(metaJson), obj.Size, sqlTime(obj.UpdatedAt))
+		`, bid, name, sqlHash256(obj.ID), sqlMD5(obj.ContentMD5), string(metaJson), obj.Size, sqlTime(obj.UpdatedAt))
 		return err
 	})
 }
 
+// UpdateCachedObject updates only the cached object metadata fields without
+// modifying the object's updated_at timestamp or other fields.
```
peterjan: If the metadata is on the `object`, why would we not update the `updated_at` column when we update the cached metadata? It's very counter intuitive to have an updated_at on a table and not update it when we update a row.

sia/persist/sqlite/init.sql
peterjan: Hm ok after doing the full review it probably does make sense to have this info on the `objects` table, but I would rename the fields.

sia/objects_test.go
```diff
@@ -648,3 +651,122 @@ func TestDeleteObjects(t *testing.T) {
 	// 	t.Fatalf("expected 0 remaining objects, got %d", *objs.KeyCount)
 	// }
 }
+
+func TestObjectMetadataCache(t *testing.T) {
```
peterjan, line 655: Extend the test cases to cover empty objects. Since we skip `sdk.Upload` for empty objects and don't have any pinned object data in those cases we should add a case for that here.

peterjan (changes requested):
This PR is a good first step but it's a little rough around the edges imo. Is there a good reason not to embed the pinned object on our `objects.Object` and just have a `GetObject` and `UpdateObject`. If the object has a cached_at we can refresh the object by fetching it from the SDK and calling `UpdateObject` with the new metadata.

sia/persist/sqlite/types.go
```diff
@@ -94,3 +97,26 @@ func (uid *sqlUploadID) Scan(src any) error {
 func (uid sqlUploadID) Value() (driver.Value, error) {
 	return uid[:], nil
 }
+
+type sqlCachedMetadata sdk.Object
+
+func (m *sqlCachedMetadata) Scan(src any) error {
+	switch src := src.(type) {
+	case string:
```
peterjan: I think it doesn't hurt adding a `[]byte` case since it's a `TEXT` field

sia/objects.go
```diff
@@ -162,6 +164,37 @@ func (s *Sia) headOrGetObject(ctx context.Context, accessKeyID *string, bucket,
 	return resp, nil
 }
 
+// getCachedObject retrieves the object metadata from cache if available and
+// fresh, otherwise fetches it from the indexer and updates the cache.
+func (s *Sia) getCachedObject(ctx context.Context, accessKeyID, bucket, object string) (sdk.Object, error) {
```
peterjan: I don't think `getCachedObject` makes sense now that the cached metadata is on the `object.Object`, right? In `headOrGetObject` we always fetch the object twice now? Instead we should check whether the metadata is valid, if not refresh the object so that in the happy path we don't have to do `GetObject` twice.

peterjan (changes requested):

sia/objects.go
```diff
@@ -164,35 +163,32 @@ func (s *Sia) headOrGetObject(ctx context.Context, accessKeyID *string, bucket,
...
-				return sdk.Object{}, fmt.Errorf("failed to fetch object from indexer: %w", err)
-			}
-		} else {
-			obj.CachedMetadata = fetched
-			obj.CachedAt = time.Now()
-			if err := s.store.PutObject(accessKeyID, bucket, object, obj); err != nil {
-				s.logger.Warn("failed to update object metadata cache", zap.Error(err))
-			}
+	// return early if cache is fresh
+	if cached && time.Since(obj.CachedAt) < metadataCacheLifetime {
```
peterjan: `nit` also an option

```
cachedUntil := obj.CachedAt.Add(metadataCacheLifetime)
if time.Now().Before(cachedUntil)
```

peterjan (dismissed):

sia/objects/objects.go
```diff
@@ -14,6 +15,9 @@ type Object struct {
 	Meta       map[string]string
 	Size       int64
 	UpdatedAt  time.Time
+
+	CachedMetadata sdk.Object // zero value if uncached (check CachedAt)
```
ChrisSchinnerl: This should always be cached right? When we upload an object we should immediately persist the metadata we get from the SDK as well.

I'd also lean towards calling this `SiaObject` or something that's not metadata. To avoid confusion with `Meta`.

sia/persist/sqlite/init.sql
```diff
@@ -18,6 +18,8 @@ CREATE TABLE objects (
     metadata TEXT NOT NULL,
     size INTEGER NOT NULL,
     updated_at INTEGER NOT NULL,
+    cached_metadata TEXT NOT NULL,
```
ChrisSchinnerl: similar thing here. Might be less confusing to have this be the `sia_object` and corresponding name for `cached_at`. 

Unless I'm missing something I'd also make this a `BLOB` since we need to seal the object before storing it and the result of that is a binary blob and not a string like `metadata`.

sia/persist/sqlite/types.go
```diff
@@ -94,3 +97,32 @@ func (uid *sqlUploadID) Scan(src any) error {
...
+			return nil
+		}
+		return json.Unmarshal(src, (*sdk.Object)(m))
+	default:
+		return fmt.Errorf("cannot scan %T to CachedMetadata", src)
+	}
+}
+
+func (m sqlCachedMetadata) Value() (driver.Value, error) {
+	data, err := json.Marshal((sdk.Object)(m))
```
ChrisSchinnerl: The object shouldn't be stored in plain text since it contains keys. We want to use the SDK to seal it before persisting it and then unseal it when reading it from disk. The sealed object supports Sia encoding.

ChrisSchinnerl (changes requested):

sia/persist/sqlite/types.go
chris124567: Can we export this then? https://github.com/SiaFoundation/indexd/blob/0a5e7770f7f961453d372ba23ba50bc5b86ac204/sdk/object.go#L179C6-L179C28

ChrisSchinnerl: Hm we definitely need to expose it somehow but I'd prefer the experience to be consistent with the Rust SDK. There we have an `open` and `seal` method on the corresponding object type.

The problem here is that `SealedObject` is declared in `slabs/objects.go` so it's a circular issue. 

What we could do is something like this
```go
type SealedObject struct {
	slabs.SealedObject
}

func (so *SealedObject) Open(appKey types.PrivateKey) (Object, error) {
	return objectFromSealedObject(*so, appKey)
}
```

and just declare a new type with an `Open` method. That gets us there in terms of consistency but embedded types aren't great. But maybe it's alright here. In that case I'd also update the whole `sdk` package to use the new type.

@n8mgr any preference?

sia/persist/sqlite/objects_test.go
peterjan, line 530: This is broken because of the NOT NULL constraint on `sia_object`

sia/persist/sqlite/multipart.go
```diff
@@ -95,16 +95,18 @@ func (s *Store) CompleteMultipartUpload(bucket, name string, uploadID s3.UploadI
 
 		// upsert object with metadata from multipart upload
 		_, err = tx.Exec(`
-			INSERT INTO objects (bucket_id, name, object_id, content_md5, metadata, size, updated_at)
-			SELECT bucket_id, name, $1, $2, metadata, $3, $4
+			INSERT INTO objects (bucket_id, name, object_id, content_md5, metadata, size, updated_at, sia_object, cached_at)
+			SELECT bucket_id, name, $1, $2, metadata, $3, $4, x'', 0
```
peterjan: This feels wrong. We have `CachedAt.IsZero` checks right? This would not pass that and likely result in false cache hits

sia/objects.go
```diff
@@ -173,6 +178,43 @@ func (s *Sia) headOrGetObject(ctx context.Context, accessKeyID *string, bucket,
...
+			// try to unseal the stale cached object
+			return s.sdk.UnsealObject(obj.SiaObject)
+		}
+		return sdk.Object{}, fmt.Errorf("failed to fetch object from indexer: %w", err)
+	}
+
+	// seal the fetched object for storage
+	obj.SiaObject = s.sdk.SealObject(fetched)
+	obj.CachedAt = time.Now()
+	if err := s.store.PutObject(accessKeyID, bucket, objectKey, obj); err != nil {
```
peterjan: Every refresh will update the objects `updated_at` field which is used for S3 LastModified, we don't want that because that would invalidate caches. We should either add a separate method to the store or pass a flag to only update the cached metadata

peterjan (changes requested):
This breaks

sia/objects.go
chris124567: I agree with you because that was what I was originally going for https://github.com/SiaFoundation/s3d/pull/95#discussion_r2727205829

sia/persist/sqlite/objects.go
```diff
@@ -128,7 +134,7 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, obj *objects.Object)
 				content_md5 = excluded.content_md5,
 				metadata = excluded.metadata,
 				size = excluded.size,
-				updated_at = excluded.updated_at,
+				updated_at = `+updatedAt+`,
```
peterjan: `nit` use `updated_at = CASE WHEN $10 THEN excluded.updated_at ELSE objects.updated_at END`

peterjan (dismissed):

sia/objects.go
peterjan: I was still young at the time

peterjan (approved):

ChrisSchinnerl (approved):

PR #101, opened 2026-03-06
Add upload packing

### LINKED ISSUE #100: Implement upload packing
S3d should make use of upload packing to keep the storage overhead for uploading lots of small files minimal.

peterjan:
This PR adds upload packing.
  
Closes https://github.com/SiaFoundation/s3d/issues/100


sia/sia.go
```diff
@@ -25,25 +25,40 @@ const (
...
-func WithKeyPair(accessKeyID, secretKey string) func(*Sia) {
-	return func(mb *Sia) {
-		mb.accessKeys[accessKeyID] = auth.SecretAccessKey(secretKey)
+func WithKeyPair(accessKeyID, secretKey string) Option {
+	return func(opts *siaOpts) {
+		opts.accessKeys[accessKeyID] = auth.SecretAccessKey(secretKey)
+	}
+}
+
+// WithPacking enables packed uploads. If not set, packing is disabled and
```
ChrisSchinnerl: I'd enable it by default. It's pretty useless without it.

sia/objects.go
```diff
@@ -265,12 +265,29 @@ func (s *Sia) PutObject(ctx context.Context, accessKeyID string, bucket, object
...
+		err = s.sdk.PinObject(ctx, obj)
+		if err != nil {
+			return nil, fmt.Errorf("failed to pin object in indexer: %w", err)
+		}
+		objectID = obj.ID()
+		siaObject = s.sdk.SealObject(obj)
+		cachedAt = time.Now()
 	} else {
-		// upload the data
+		// large object or unknown size — upload directly
```
ChrisSchinnerl: I'm pretty sure S3 requires the content length to always be set. Otherwise `putObject` will fail with a `ErrMissingContentLength` error. Which is probably also why `PutObjectOptions` uses a `int64` rather than `*int64`  for the `ContentLength`.

sia/packing.go
```diff
@@ -0,0 +1,252 @@
...
+	"context"
+	"errors"
+	"io"
+	"sync"
+	"time"
+
+	"go.sia.tech/indexd/sdk"
+)
+
+const flushTimeout = time.Second
```
ChrisSchinnerl: Flushing after some time has passed reduces the usefulness of upload packing in s3d significantly. If my use case requires updating a small file every 60 seconds or so I'll end up with crazy amounts of overhead.

Considering that S3D only needs to work on one server and not sync across phones etc. it should be acceptable to store small files on disk and only when you have enough data to upload efficiently, upload the object with an upload packer and save it to Sia. Before then it can be served from disk instead. 

s3d is inherently centralized since clients don't use the Sia network directly so we might as well lean into that to make things a bit nicer.

ChrisSchinnerl (changes requested):

sia/packing.go
peterjan: Oof... that changes things 😅. I'll refactor it.

sia/objects_test.go
```diff
@@ -147,6 +149,191 @@ func TestGetAndHeadObject(t *testing.T) {
 	}
 }
 
+func TestObjectPacking(t *testing.T) {
```
peterjan: merge blocker, reminder to do more extensive manual tests

cmd/s3d/main.go
```diff
@@ -208,10 +205,22 @@ func main() {
 		checkFatalError("failed to get app key from database", err)
 	}
 
-	backend, err := sia.New(ctx, sia.NewSDK(sdkClient), store, cfg.Directory, sia.WithKeyPair(cfg.Sia.AccessKey, cfg.Sia.SecretKey))
+	// add key pair from env vars if set
+	if ak, sk := os.Getenv(accessKeyEnv), os.Getenv(secretKeyEnv); ak != "" && sk != "" {
```
ChrisSchinnerl: I think we can get rid of those env vars altogether. I'd use the same approach here as we do in indexd. If there is no reason why something needs to be a CLI flag or env var, only use the config file.
Usually the only exception is whatever you need to make it run within a docker container.

sia/persist/sqlite/objects.go
```diff
@@ -70,6 +74,7 @@ func (s *Store) GetObject(accessKeyID *string, bucket, name string, partNumber *
 		return nil, err
 	}
 
+	obj.Name = name
```
ChrisSchinnerl: It's not obvious that this is necessary after all successful call to `getObject`. Better to move that into `getObject`.

sia/packing.go
```diff
@@ -0,0 +1,266 @@
...
+	"go.sia.tech/indexd/sdk"
+	"go.uber.org/zap"
+	"lukechampine.com/frand"
+)
+
+const (
+	// DefaultPackingThreshold is the slab size at default redundancy (40 MiB).
+	// objects smaller than this are stored on disk until enough data has
+	// accumulated to fill a slab.
+	DefaultPackingThreshold = 40 << 20
```
ChrisSchinnerl: This constant is obsolete. We can use the slab size computed from the redundancy settings the user passed in. 

e.g. if we have 400MiB slabs we don't want to use 40MiB as a threshold because that's just wasteful but we also don't want users to think about slabs at all. Better to just give them a percentage to tweak. e.g. 10% waste.

sia/objects.go
```diff
@@ -248,17 +318,35 @@ func (s *Sia) PutObject(ctx context.Context, accessKeyID string, bucket, object
...
 	var cachedAt time.Time
+	var filename *string
 	if opts.ContentLength == 0 {
 		// drain reader
 		if _, err := io.Copy(io.Discard, lr); err != nil {
 			return nil, fmt.Errorf("failed to read object data: %w", err)
 		}
+	} else if s.needsPacking(opts.ContentLength) {
+		// small object, write to disk and upload to Sia later
+		fn, err := s.writeToDisk(lr)
```
ChrisSchinnerl: The `needsPacking` check isn't very efficient. A large file can still produce a lot of waste.

e.g. 20MiB at 40MiB slab size is 50% waste. But 60MiB out of 80MiB is still 25% waste. Which is greater than the 10% leeway.

So I'd update `needsPacking` to compute the waste in percent regardless of file size and only ignore packing if it's < the 10% waste threshold. 

That will also result in a natural limit for files on disk depending on the threshold. At 10% it should be 360 MiB at default redundancy.

sia/packing.go
```diff
@@ -0,0 +1,266 @@
...
+		s.tryRemove(&obj.Filename)
+		s.logger.Debug("packed object uploaded to Sia",
+			zap.String("bucket", obj.Bucket),
+			zap.String("name", obj.Name))
+	}
+
+	return true, nil
+}
+
+func (s *Sia) tryPack(filename *string) {
```
ChrisSchinnerl: This is currently only run on demand. If I upload a bunch of files, then restart and then don't upload, the existing files won't get packed even if they exceed the threshold.

Same is true if I stop s3d, update the threshold to something higher, new files are now ready to get packed but it won't trigger without an upload.

Might be better to run a background loop which is both triggered by uploads and checks every minute or so by itself.

sia/sia.go
```diff
@@ -24,22 +26,39 @@ const (
...
+	return func(s *Sia) {
+		s.accessKeys[accessKeyID] = auth.SecretAccessKey(secretKey)
+	}
+}
+
+// WithPacking sets the packing threshold and leeway percentage. Objects smaller than
+// the threshold are stored on disk until enough data has accumulated to upload
+// efficiently. The leeway percentage determines how much smaller the remaining space
+// in a slab can be compared to the threshold before packing is triggered. Pass a 0
+// threshold to disable packing.
```
ChrisSchinnerl: I'd condense those 2 values into a single threshold which is the percentage of waste tolerable and defaults to 10%.

sia/packing.go
```diff
@@ -0,0 +1,266 @@
...
+				zap.Int64("threshold", s.packingThreshold))
+			break
+		}
+
+		// pack a slab
+		s.logger.Info("packing slab",
+			zap.Int("candidates", len(objs)),
+			zap.Int64("totalSize", totalSize))
+
+		if packed, err := s.packSlab(ctx, objs); err != nil {
```
ChrisSchinnerl: This is going to be quite slow considering that we limit ourselves to 1 packed upload at a time. 

We could separate the logic for picking packed objects and actually uploading them and then run 4 or so uploads in parallel.

sia/packing.go
```diff
@@ -0,0 +1,266 @@
...
+			s.logger.Error("failed to pack slab", zap.Error(err))
+			return
+		} else if !packed {
+			s.logger.Debug("no objects were packed into the slab")
+			break
+		}
+	}
+}
+
+func (s *Sia) packSlab(ctx context.Context, candidates []objects.PackedObject) (bool, error) {
```
ChrisSchinnerl: (nit) `packObjects`? Don't think we need to mention Slabs at all.

sia/persist/sqlite/objects.go
```diff
@@ -108,35 +113,48 @@ func getObject(tx *txn, obj *objects.Object, bid int64, name string, partNumber
...
+		return nil, err
+	}
+	return prevFilename, nil
 }
 
 // CopyObject atomically reads the source object and writes it to the
 // destination within a single transaction, applying metadata according to the
 // replace flag. Returns the copied object metadata.
-func (s *Store) CopyObject(srcBucket, srcName, dstBucket, dstName string, meta map[string]string, replace bool) (result *objects.Object, err error) {
+func (s *Store) CopyObject(srcBucket, srcName, dstBucket, dstName string, meta map[string]string, replace bool, dstFilename *string) (_ *objects.Object, prevFilename *string, _ error) {
```
ChrisSchinnerl: It's a shame to copy the buffered file but maybe it's fine for a start. I don't expect a lot of copying to happen in general.

ChrisSchinnerl (changes requested):

sia/persist/sqlite/objects.go
peterjan: Yeah if it turns out to be super wasteful we can improve it

sia/packing.go
peterjan: Ok, I considered this when writing it of course but decided against it because I don't really think we needed a background loop per se. The above will almost never be true in any practical use case, and how someone will run s3d it'll be perfectly fine for the files to be on disk or made it on to Sia. I'll add it though

sia/objects.go
```diff
@@ -28,10 +30,49 @@ const metadataCacheLifetime = 24 * time.Hour
...
+	// if the source is on disk, copy the file first so we can pass the
+	// new filename to the store
+	var dstFilename *string
+	srcObj, err := s.store.GetObject(&accessKeyID, srcBucket, srcObject, nil)
 	if err != nil {
 		return nil, err
 	}
+	if srcObj.Filename != nil {
+		src, err := os.Open(filepath.Join(s.packingDir, *srcObj.Filename))
+		if errors.Is(err, os.ErrNotExist) {
```
ChrisSchinnerl: (nit) the file opening logic could be extracted into a helper accepting the object itself that returns either a valid file handle and nil error, no file handle and nil error or no file handle and an error.

sia/persist/sqlite/init.sql
ChrisSchinnerl: It's a bit odd that `filename` is nullable but `object_id` is not and uses a default value. Considering that `object_id` is also optional now it might be better to have the database reflect that and maybe add a `CHECK` constraint that guarantees only one of the two columns is NULL and the other one is set.

sia/sdk.go
```diff
@@ -70,6 +70,28 @@ func (s *IndexdSDK) Upload(ctx context.Context, r io.Reader) (sdk.Object, error)
 	return obj, nil
 }
 
+// SlabSize returns the slab size by creating a temporary packed upload and
+// reading its capacity.
+func (s *IndexdSDK) SlabSize() (int64, error) {
+	pu, err := s.inner.UploadPacked(s.ulOpts...)
```
ChrisSchinnerl, line 64: Just expose the `SlabSize` on the packed upload like we do in the Rust SDK. This is a pretty awful workaround.

sia/sia.go
```diff
@@ -88,29 +115,73 @@ type Store interface {
...
+		if err != nil {
+			return nil, err
+		}
+		go func() {
+			defer cancel()
+			s.packingLoop(ctx)
+		}()
+	}
+
+	// TODO: add loop to cleanup orphaned packed objects on disk
```
ChrisSchinnerl: Either create an issue for a followup or address it here. This will just end up in the void.

sia/packing.go
```diff
@@ -0,0 +1,420 @@
...
+// current slab's remaining space.
+func (p *packedObjects) fits(obj objects.PackedObject) bool {
+	if p.totalSize+obj.Length > p.slabSize*packedUploadMaxSlabs {
+		return false
+	}
+	return obj.Length <= p.slabRemaining() || obj.Length > p.slabSize
+}
+
+// add appends the object to this group.
+func (p *packedObjects) add(obj objects.PackedObject) {
```
ChrisSchinnerl: I think we can actually simplify this quite a bit:

1. the order the database should return the packable objects from smallest -> largest
2. we get rid of `fits`
3. `add` returns a bool with either true or false depending on whether an object was added

Then all you need to do is loop over the packable objects once and for each group call `add` until it returns `false`. 

The logic in `add` is the following
- if the group itself is not below the waste threshold, add the object
- if the group itself is below the waste threshold, add the object if it fits into the remaining space

Thanks to the ascending order, as soon as `add` returns `false` for the first time we know that the group is finished because if the current object didn't fit, the next one also won't fit. 

In theory you also don't need the `packedObjects` type anymore because you will now cleanly partition the packable objects very efficiently. Only scanning all of the objects once.

Regarding the `packedUploadMaxSlabs`, we don't need that anymore because the waste threshold enforces a natural upper limit. The more objects you add, the smaller the waste eventually.

sia/packing.go
```diff
@@ -0,0 +1,420 @@
...
+		return
+	}
+	if err := os.Remove(filepath.Join(s.packingDir, *filename)); err != nil && !errors.Is(err, os.ErrNotExist) {
+		s.logger.Error("failed to remove file on disk",
+			zap.String("filename", *filename),
+			zap.Error(err))
+	}
+}
+
+func (s *Sia) writeToDisk(r io.Reader) (filename string, err error) {
```
ChrisSchinnerl: The naming of these methods is a bit generic. I'd change it to make it clear that they are related to writing and removing packed objects.

sia/packing.go
```diff
@@ -0,0 +1,420 @@
...
+			s.logger.Debug("packing triggered")
+			t.Reset(5 * time.Minute)
+		case <-t.C:
+		}
+
+		s.packObjects(ctx)
+	}
+}
+
+func (s *Sia) packObjects(ctx context.Context) {
```
ChrisSchinnerl: Not sure I understand the benchmark in the description but overall the throughput doesn't seem to spark joy. Can you check the throughput you get when uploading about 1-5GB of tiny files? 
I'd expect the throughput to disk to saturate 1Gbps easily but really the packing should also be fast.

ChrisSchinnerl (changes requested):

sia/persist/sqlite/init.sql
peterjan: I considered that but I was not looking forward to making ID a pointer on the objects struct. The `CHECK` is not that straightforward neither because of empty objects (both are NULL and size is 0). I made object ID a pointer though and added this constraint `CHECK(object_id IS NULL OR filename IS NULL)` which guarantees a file can never be on disk and stored on Sia at the same time.

sia/packing.go
peterjan: Ok I ran some tests:

Upload throughput for uploading ~5GB in small files (1kb to 10mb) seems to be very consistent at ~250MB/s. Quite a bit slower than I expected but perhaps it's just my crappy hardware. 

```
=== UPLOAD THROUGHPUT ===
  Data:       5123.71 MB (1024 files)
  Time:       22.36 s
  Throughput: 229.11 MB/s
```
  
If I run nload during packing I can see it averages around 5-600Mbit/s. I added some logging of course and found that throughput is roughly 20-25MB/s. I also found that increasing the number of threads, from 4->8 doesn't meaningfully increase throughput, and the same goes for max slabs, which I bumped from 4->8 as well with no meaningful changes.

The packing is near perfect at an average of 0.05% waste (so near perfect slabs), which is to be expected for a bunch of small files where a single file never meets the threshold. Using files greater than a slab size I found waste generally averages out around 2%, which is still great imo.

sia/packing.go
peterjan: Hm that feels very counter intuitive to me. A large object is hard to fit and a small object is easy to fit. Therefor you should fit larger objects first and fill the gaps with small objects. I briefly researched this before implementing it and it's essentially a bin packing problem right. From what I've read online the current approach is called First Fit Decreasing and it's generally a good default heuristic with a good complexity trade-off. 

I don't mind playing around with the logic and doing some tests, but considering we have near perfect packing, and considering the above method is hardly researched and the consensus online seems to be it's a sub-optimal approach.. should we keep the current implementation?

I updated and simplified it to get rid of the nested loop. The code looks a lot cleaner now I think. Perhaps this version wouldn't have triggered you.

sia/sdk.go
peterjan: I'll open a PR for that, even with that method exposed it still feels awful and we still need to work around it. Perhaps we should consider specifying the redundancy on the SDK level and remove it from the upload options. That would get rid of some friction imo, even though it also feels a little weird.

sia/persist/sqlite/init.sql
```diff
@@ -13,16 +13,19 @@ CREATE TABLE buckets (
...
-    object_id BLOB NOT NULL,
+    object_id BLOB,
     content_md5 BLOB NOT NULL,
     metadata TEXT NOT NULL,
     size INTEGER NOT NULL,
     updated_at INTEGER NOT NULL,
     sia_object BLOB NOT NULL,
     cached_at INTEGER NOT NULL,
+    filename TEXT,
+    CHECK(object_id IS NULL OR filename IS NULL),
```
ChrisSchinnerl:
```SQL
CHECK((object_id IS NULL AND filename IS NOT NULL) OR (object_id IS NOT NULL AND filename IS NULL) OR size = 0), -- object is either uploaded on Sia, stored on disk or has no content
```

Would be a bit more specific

sia/persist/sqlite/multipart.go
```diff
@@ -102,33 +104,36 @@ func (s *Store) CompleteMultipartUpload(bucket, name string, uploadID s3.UploadI
...
-		}
-
-		if oldID != nil && *oldID != objectID {
-			if err := insertOrphan(tx, *oldID); err != nil {
+		if objectID != nil {
+			if _, err := tx.Exec("DELETE FROM orphaned_objects WHERE object_id = $1", sqlHash256(*objectID)); err != nil {
 				return err
 			}
+
+			if oldID != nil && *oldID != *objectID {
```
ChrisSchinnerl: Make sure to extend the orphan testing because this is currently incorrect for the case where `objectID` is `nil` but `oldID` isn't. You still orphan the old id but fail to insert the orphan.

So this should be moved out of the outer `if` and extended to check `if oldID != nil && (objectID == nil || *oldID != *objectID)`

sia/persist/sqlite/objects.go
```diff
@@ -199,58 +205,72 @@ func (s *Store) RemoveOrphanedObject(objectID types.Hash256) error {
...
-		return err
+	if obj.ID != nil {
+		if _, err := tx.Exec("DELETE FROM orphaned_objects WHERE object_id = $1", sqlHash256(*obj.ID)); err != nil {
+			return nil, err
+		}
 	}
 
-	if oldID != nil && *oldID != obj.ID {
-		return insertOrphan(tx, *oldID)
+	if oldID != nil && (obj.ID == nil || *oldID != *obj.ID) {
```
ChrisSchinnerl: here it seems to be correct

sia/persist/sqlite/types.go
```diff
@@ -30,6 +30,9 @@ type sqlHash256 types.Hash256
 
 func (h *sqlHash256) Scan(src any) error {
 	switch src := src.(type) {
+	case nil:
```
ChrisSchinnerl: Better to introduce the `nullable` type we have in indexd and be explicit than doing this.

sia/persist/sqlite/types.go
```diff
@@ -42,9 +45,27 @@ func (h *sqlHash256) Scan(src any) error {
 }
 
 func (h sqlHash256) Value() (driver.Value, error) {
+	if h == (sqlHash256{}) {
+		return nil, nil
+	}
 	return h[:], nil
 }
 
+func (h sqlHash256) Ptr() *types.Hash256 {
```
ChrisSchinnerl: You don't need this if you use casting and `nullable`.

sia/sdk.go
ChrisSchinnerl: This can be updated now

sia/packing.go
```diff
@@ -0,0 +1,456 @@
...
+)
+
+const (
+	// DefaultPackingWastePct is the maximum percentage of wasted space that is tolerated
+	// before an object is packed. Objects whose upload would waste more than this percentage are
+	// written to disk and batched together with other small objects to fill slabs efficiently.
+	DefaultPackingWastePct = 0.1
+
+	packedUploadThreads  = 6
+	packedUploadMaxSlabs = 4
```
ChrisSchinnerl: You can drop this arbitrary limit and instead check whether or not the upload meets the upload threshold within `tryAdd`. If it doesn't, allow objects > slabSize and if it does, only allow objects that fit in the remaining space.

One less constant and slightly larger groups which might result in less waiting time on slow hosts since more slabs get interleaved.

sia/packing.go
```diff
@@ -0,0 +1,456 @@
...
+			zap.Float64("wastePct", g.wastePct()))
+	}
+
+	return ready
+}
+
+// preparePendingUploads creates packed uploads for the given
+// packs, reading files from disk into each upload. Packs
+// whose files have been deleted concurrently are discarded.
+func (s *Sia) preparePendingUploads(ctx context.Context, packs []packedObjects) (_ []pendingUpload, err error) {
```
ChrisSchinnerl: This name is a bit misleading because this is actually uploading already before even spawning the threads for uploading. Which is something you probably want to fix.

sia/packing.go
ChrisSchinnerl: I'm not too surprised that more threads doesn't increase throughput because `PackedUpload.Add` is blocking already and you do that before spawning the goroutines it seems like. Only `Finalize` happens in the goroutine but at that point the whole packed object is already done uploading for the most part.

ChrisSchinnerl (changes requested):

sia/sdk.go
peterjan: I've already updated it. I have the feeling there's a misunderstanding here maybe? We still have to create a packed upload and insta close it to get at the slab size.

sia/packing.go
peterjan: Thanks for catching that

peterjan:
@ChrisSchinnerl I ran some tests and uploading the files is quite slow at ~200 MiB/s but once packing kicks in I can see a very constant upload rate of ~950Mbit/s. I can't really get the average above 650 though but I've been testing with 3-5GB runs. There's no drops in outgoing traffic so a larger test should saturate my connection fully.

I ran several configurations, with various file sizes. If you go with tiny files we fill slabs almost perfectly and upload single slab objects. If you go with random files between 1kb and a little over the slab size you get very different results though, which is to be expected obviously. Waste in general is below 1% even though the threshold is at 10% so we do a good job at minimising it.

sia/persist/sqlite/types.go
```diff
@@ -45,6 +46,38 @@ func (h sqlHash256) Value() (driver.Value, error) {
 	return h[:], nil
 }
 
+type sqlNullHash256 struct {
```
ChrisSchinnerl: You can drop all of this if you do the following:

```
diff --git a/sia/persist/sqlite/multipart.go b/sia/persist/sqlite/multipart.go
index 049bd33..39ff880 100644
--- a/sia/persist/sqlite/multipart.go
+++ b/sia/persist/sqlite/multipart.go
@@ -117,7 +117,7 @@ func (s *Store) CompleteMultipartUpload(bucket, name string, uploadID s3.UploadI
 				sia_object = excluded.sia_object,
 				cached_at = excluded.cached_at,
 				filename = excluded.filename
-		`, sqlNullableHash256(&objectID), sqlMD5(contentMD5), contentLength, sqlTime(time.Now()), sqlTime(time.Time{}), filename, sqlUploadID(uploadID))
+		`, (*sqlHash256)(objectID), sqlMD5(contentMD5), contentLength, sqlTime(time.Now()), sqlTime(time.Time{}), filename, sqlUploadID(uploadID))
 		if err != nil {
 			return err
 		}
diff --git a/sia/persist/sqlite/objects.go b/sia/persist/sqlite/objects.go
index 4d9ad8c..6e9cb5b 100644
--- a/sia/persist/sqlite/objects.go
+++ b/sia/persist/sqlite/objects.go
@@ -28,14 +28,14 @@ func (s *Store) DeleteObject(accessKeyID, bucket string, objectID s3.ObjectID) (
 
 		// delete the row and return its values for precondition checks and
 		// orphan detection; the transaction rolls back if preconditions fail
-		var deletedID *types.Hash256
+		var deletedID sql.Null[sqlHash256]
 		var contentMD5 [16]byte
 		var size int64
 		var updatedAt time.Time
 		err = tx.QueryRow(`
 			DELETE FROM objects WHERE bucket_id = $1 AND name = $2
 			RETURNING object_id, content_md5, size, updated_at, filename
-		`, bid, objectID.Key).Scan(sqlNullableHash256(&deletedID), (*sqlMD5)(&contentMD5), &size, (*sqlTime)(&updatedAt), &filename)
+		`, bid, objectID.Key).Scan(&deletedID, (*sqlMD5)(&contentMD5), &size, (*sqlTime)(&updatedAt), &filename)
 		if errors.Is(err, sql.ErrNoRows) {
 			return nil
 		} else if err != nil {
@@ -52,8 +52,8 @@ func (s *Store) DeleteObject(accessKeyID, bucket string, objectID s3.ObjectID) (
 			return s3errs.ErrPreconditionFailed
 		}
 
-		if deletedID != nil {
-			return insertOrphan(tx, *deletedID)
+		if deletedID.Valid {
+			return insertOrphan(tx, types.Hash256(deletedID.V))
 		}
 		return nil
 	}); err != nil {
@@ -101,11 +101,13 @@ func getObject(tx *txn, obj *objects.Object, bucket, name string, partNumber *in
 			obj.PartsCount = *partNumber
 		}
 		var siaObj sqlSiaObject
+		var objID sql.Null[*sqlHash256]
 		err := tx.QueryRow(`
 			SELECT object_id, metadata, updated_at, size, content_md5, sia_object, cached_at, filename
 			FROM objects
-		`, bid, name).Scan(sqlNullableHash256(&obj.ID), (*sqlMetaJSON)(&obj.Meta), (*sqlTime)(&obj.LastModified), &obj.Length, (*sqlMD5)(&obj.ContentMD5), &siaObj, (*sqlTime)(&obj.CachedAt), &obj.Filename)
+		`, bid, name).Scan(&objID, (*sqlMetaJSON)(&obj.Meta), (*sqlTime)(&obj.LastModified), &obj.Length, (*sqlMD5)(&obj.ContentMD5), &siaObj, (*sqlTime)(&obj.CachedAt), &obj.Filename)
+		obj.ID = (*types.Hash256)(objID.V)
 		obj.SiaObject = slabs.SealedObject(siaObj)
 		obj.Name = name
 		obj.Bucket = bucket
@@ -119,12 +121,14 @@ func getObject(tx *txn, obj *objects.Object, bucket, name string, partNumber *in
 
 	// part specified, return part info
 	var siaObj sqlSiaObject
+	var objID sql.Null[*sqlHash256]
 	err = tx.QueryRow(`
 		SELECT o.object_id, o.metadata, o.updated_at, o.sia_object, o.cached_at, o.filename, p.offset, p.content_length, p.content_md5
 		FROM object_parts p
 		JOIN objects o ON o.bucket_id = p.bucket_id AND o.name = p.name
 		WHERE o.bucket_id = $1 AND o.name = $2 AND p.part_number = $3
-	`, bid, name, *partNumber).Scan(sqlNullableHash256(&obj.ID), (*sqlMetaJSON)(&obj.Meta), (*sqlTime)(&obj.LastModified), &siaObj, (*sqlTime)(&obj.CachedAt), &obj.Filename, &obj.Offset, &obj.Length, (*sqlMD5)(&obj.ContentMD5))
+	`, bid, name, *partNumber).Scan(&objID, (*sqlMetaJSON)(&obj.Meta), (*sqlTime)(&obj.LastModified), &siaObj, (*sqlTime)(&obj.CachedAt), &obj.Filename, &obj.Offset, &obj.Length, (*sqlMD5)(&obj.ContentMD5))
+	obj.ID = (*types.Hash256)(objID.V)
 	obj.SiaObject = slabs.SealedObject(siaObj)
 	obj.Name = name
 	obj.Bucket = bucket
@@ -231,7 +235,7 @@ func putObject(tx *txn, bucket, name string, obj objects.Object, updateModTime b
 			sia_object = excluded.sia_object,
 			cached_at = excluded.cached_at,
 			filename = excluded.filename
-	`, bID, name, sqlNullableHash256(&obj.ID), sqlMD5(obj.ContentMD5),
+	`, bID, name, (*sqlHash256)(obj.ID), sqlMD5(obj.ContentMD5),
 		sqlMetaJSON(obj.Meta), obj.Length, sqlTime(time.Now()),
 		sqlSiaObject(obj.SiaObject), sqlTime(obj.CachedAt), obj.Filename, updateModTime)
 	if err != nil {
@@ -255,7 +259,7 @@ func putObject(tx *txn, bucket, name string, obj objects.Object, updateModTime b
 // objectInfo returns the object_id and filename currently stored
 // for the given bucket and name, or nils if no row exists.
 func objectInfo(tx *txn, bucket, name string) (*types.Hash256, *string, error) {
-	var id *types.Hash256
+	var objID sql.Null[*sqlHash256]
 	var filename *string
 	err := tx.
 		QueryRow(`
@@ -263,13 +267,13 @@ func objectInfo(tx *txn, bucket, name string) (*types.Hash256, *string, error) {
 			FROM objects o
 			INNER JOIN buckets b ON b.id = o.bucket_id
 			WHERE b.name = $1 AND o.name = $2`, bucket, name).
-		Scan(sqlNullableHash256(&id), &filename)
+		Scan(&objID, &filename)
 	if errors.Is(err, sql.ErrNoRows) {
 		return nil, nil, nil
 	} else if err != nil {
 		return nil, nil, err
 	}
-	return id, filename, nil
+	return (*types.Hash256)(objID.V), filename, nil
 }
 
 // insertOrphan adds objectID to the orphaned_objects table if no rows in the
diff --git a/sia/persist/sqlite/types.go b/sia/persist/sqlite/types.go
index 5fdaeb5..7fe2aac 100644
--- a/sia/persist/sqlite/types.go
+++ b/sia/persist/sqlite/types.go
@@ -14,7 +14,6 @@ import (
 
 var (
 	_ scannerValuer = (*sqlHash256)(nil)
-	_ scannerValuer = (*sqlNullHash256)(nil)
 	_ scannerValuer = (*sqlMD5)(nil)
 	_ scannerValuer = (*sqlTime)(nil)
 	_ scannerValuer = (*sqlUploadID)(nil)
@@ -46,38 +45,6 @@ func (h sqlHash256) Value() (driver.Value, error) {
 	return h[:], nil
 }
 
-type sqlNullHash256 struct {
-	h **types.Hash256
-}
-
-func sqlNullableHash256(h **types.Hash256) *sqlNullHash256 {
-	return &sqlNullHash256{h: h}
-}
-
-func (h *sqlNullHash256) Scan(src any) error {
-	switch src := src.(type) {
-	case nil:
-		*h.h = nil
-		return nil
-	case []byte:
-		if len(src) != len(types.Hash256{}) {
-			return fmt.Errorf("failed to scan source into nullable Hash256 due to invalid number of bytes %v != %v: %v", len(src), len(types.Hash256{}), src)
-		}
-		*h.h = new(types.Hash256)
-		copy((*h.h)[:], src)
-		return nil
-	default:
-		return fmt.Errorf("cannot scan %T to nullable Hash256", src)
-	}
-}
-
-func (h sqlNullHash256) Value() (driver.Value, error) {
-	if *h.h == nil {
-		return nil, nil
-	}
-	return (*h.h)[:], nil
-}
-
 type sqlTime time.Time
 
 func (t *sqlTime) Scan(src any) error {
```

ChrisSchinnerl (commented):
Looking good now but I have looked at this so many times it would be good to have a fresh pair of eyes take a look too.

@chris124567 could you give this a review as well

chris124567:
@ChrisSchinnerl I will take a look

sia/persist/sqlite/multipart.go
```diff
@@ -89,7 +90,8 @@ func (s *Store) CompleteMultipartUpload(bucket, name string, uploadID s3.UploadI
 			return fmt.Errorf("found %d parts smaller than minimum size (%d bytes)", smallParts, s3.MinUploadPartSize)
 		}
 
-		oldID, err := previousObjectID(tx, bid, name)
+		oldID, oldFilename, err := objectInfo(tx, bucket, name)
```
chris124567: You could change `objectInfo` to take `bid` directly

sia/objects.go
```diff
@@ -245,21 +297,41 @@ func (s *Sia) PutObject(ctx context.Context, accessKeyID string, bucket, object
...
+		packedFilename = &filename
 	} else {
-		// upload the data
+		// upload directly and pin the object
 		obj, err := s.sdk.Upload(ctx, lr)
 		if err != nil {
 			return nil, fmt.Errorf("failed to upload object: %w", err)
 		}
-		objectID = obj.ID()
+		err = s.sdk.PinObject(ctx, obj)
```
chris124567: I think we don't need to pin again because sdk.Upload should pin already

sia/objects.go
```diff
@@ -28,11 +28,40 @@ const metadataCacheLifetime = 24 * time.Hour
...
+		defer file.Close()
+		fn, err := s.writePackedObject(file)
+		if err != nil {
+			return nil, fmt.Errorf("failed to copy packed object: %w", err)
+		}
+		dstFilename = &fn
+	}
+
+	// copy the object in the store
+	obj, prevFilename, err := s.store.CopyObject(srcBucket, srcObject, dstBucket, dstObject, meta, replace, dstFilename)
```
chris124567: If an object has been finalized somewhere between the GetObject call above and here, won't this CopyObject call fail because we have the CHECK in the schema that makes sure only either the filename or object_id can be set?

sia/objects.go
peterjan: The SDK doesn't pin the object on Upload. Edit: oh wait we wrap that, heh that's a little confusing, will update

sia/objects.go
peterjan: Damn, nice catch sir.

peterjan:
### Before (old `tryAdd` logic)                                                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                      
  | # | Objects | Size (MB) | Slabs | Waste |                                                                                                                                                                                                                                                                                                                                                                         
  |---|---------|-----------|-------|-------|                                                                                                                                                                                                                                                                                                                                                                         
  | 1 | 5 | 159.8 | 4 | 0.13% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 2 | 6 | 199.9 | 5 | 0.05% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 3 | 5 | 199.8 | 5 | 0.10% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 4 | 6 | 239.6 | 6 | 0.14% |                                   
  | 5 | 6 | 239.5 | 6 | 0.20% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 6 | 9 | 359.9 | 9 | 0.02% |           
  | 7 | 2 | 39.6 | 1 | 0.94% |                                                                                                                                                                                                                                                                                                                                                                                        
  | 8 | **1** | 38.2 | 1 | **4.57%** |                                                                                                                                                                                                                                                                                                                                                                                
  | 9 | **1** | 38.1 | 1 | **4.76%** |                                                                                                                                                                                                                                                                                                                                                                                
  | 10 | 2 | 39.7 | 1 | 0.74% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 11 | **1** | 37.1 | 1 | **7.29%** |                                                                                                                                                                                                                                                                                                                                                                               
  | 12 | **1** | 36.8 | 1 | **8.11%** |                                                                                                                                                                                                                                                                                                                                                                               
  | 13 | **1** | 36.6 | 1 | **8.51%** |                                                                                                                                                                                                                                                                                                                                                                               
  | 14 | **1** | 36.3 | 1 | **9.20%** |                                                                                                                                                                                                                                                                                                                                                                               
  | 15 | 2 | 39.9 | 1 | 0.13% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 16 | 7 | 238.7 | 6 | 0.55% |                                                                                                                                                                                                                                                                                                                                                                                      
  | 17 | 7 | 199.8 | 5 | 0.12% |                                                                                                                                                                                                                                                                                                                                                                                      
  | 18 | 4 | 117.1 | 3 | 2.37% |                                                                                                                                                                                                                                                                                                                                                                                      
  | 19 | 5 | 118.4 | 3 | 1.29% |                                                                                                                                                                                                                                                                                                                                                                                      
  | 20 | 5 | 119.2 | 3 | 0.65% |                                                                                                                                                                                                                                                                                                                                                                                      
  | 21 | 4 | 79.3 | 2 | 0.84% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 22 | 4 | 79.3 | 2 | 0.84% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 23 | 8 | 158.9 | 4 | 0.69% |                                                                                                                                                                                                                                                                                                                                                                                      
  | 24 | 8 | 116.4 | 3 | 2.98% |                                                                                                                                                                                                                                                                                                                                                                                      
  | 25 | 6 | 79.1 | 2 | 1.13% |                                                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                      
  ### After (new `tryAdd` logic with `maxSlabsPerPack=6`)                                                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                      
  | # | Objects | Size (MB) | Slabs | Waste |                                                                                                                                                                                                                                                                                                                                                                         
  |---|---------|-----------|-------|-------|                                                                                                                                                                                                                                                                                                                                                                         
  | 1 | 6 | 238.9 | 6 | 0.42% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 2 | 6 | 239.8 | 6 | 0.06% |                                                                                                                                                                                                                                                                                                                                                                                     
  | 3 | 6 | 239.5 | 6 | 0.21% |                                   
  | 4 | 6 | 239.4 | 6 | 0.22% |                 
  | 5 | 7 | 239.7 | 6 | 0.13% |           
  | 6 | 7 | 239.9 | 6 | 0.02% |               
  | 7 | 8 | 239.5 | 6 | 0.19% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 8 | 7 | 199.8 | 5 | 0.08% |
  | 9 | 6 | 119.6 | 3 | 0.36% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 10 | 4 | 79.9 | 2 | 0.08% |                                                                                                                                                                                                                                                                                                                                                                                     
  | 11 | 4 | 79.8 | 2 | 0.26% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 12 | 6 | 119.9 | 3 | 0.06% |                                                                                                                                                                                                                                                                                                                                                                                    
  | 13 | 6 | 119.2 | 3 | 0.66% |                                  
  | 14 | 8 | 158.7 | 4 | 0.84% |                                                                                                                                                                                                                                                                                                                                                                                      
  | 15 | 4 | 79.3 | 2 | 0.88% |           
  | 16 | 2 | 38.0 | 1 | 5.06% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 17 | 10 | 159.4 | 4 | 0.34% |                                                                                                                                                                                                                                                                                                                                                                                     
  | 18 | 5 | 78.9 | 2 | 1.36% |                                   
  | 19 | 7 | 79.1 | 2 | 1.06% |                                                                                                                                                                                                                                                                                                                                                                                       
  | 20 | 4 | 38.4 | 1 | 4.08% |                                                                                                                                                                                                                                                                                                                                                                                     
  | 21 | 5 | 39.8 | 1 | 0.41% |                                                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                      
  ### Summary                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                      
  |  | Before | After | Change |                                                                                                                                                                                                                                                                                                                                                                                    
  |--|--------|-------|--------|                                                                                                                                                                                                                                                                                                                                                                                      
  | Objects packed | 111 | 130 | +19 |                                                                                                                                                                                                                                                                                                                                                                              
  | Packs | 25 | 21 | 4 fewer |                                                                                                                                                                                                                                                                                                                                                                                       
  | Total slabs | 78 | 83 | +5 (more objects) | 
  | Max waste | 9.20% | 5.06% | down 45% |                                                                                                                                                                                                                                                                                                                                                                            
  | Packs with >4% waste | 6 | 2 | down 67% |                                                                                                                                                                                                                                                                                                                                                                       
  | Single-object packs | 6 | 0 | eliminated |                    
  | Leftover objects | 4 (stuck) | 6 | similar |                                                                                                                                                                                                                                                                                                                                                                      
  | Max slabs per pack | 9 | 6 | capped |
  
  @ChrisSchinnerl I want to update the packing algorithm and re-introduce `maxSlabsPerPack`. Even though it's an arbitrary number it's meant to ensure we get all benefits from parallelising the uploads, as well as prevent single slab objects with high waste pct. It's a minor (so cheap) change with generally a lot better results so we should add it back in (?)

sia/persist/sqlite/objects.go
```diff
@@ -161,8 +161,9 @@ func (s *Store) PutObject(accessKeyID, bucket, name string, contentMD5 [16]byte,
 
 // MarkObjectUploaded transitions a pending upload to a Sia-backed object by
 // setting the object_id, sia_object and cached_at fields while clearing
-// filename. Fails if the object already has an object_id.
-func (s *Store) MarkObjectUploaded(bucket, name string, siaObject slabs.SealedObject) error {
+// filename. The update only proceeds if the current filename matches
+// expectedFilename, returning ErrObjectModified when zero rows are affected.
+func (s *Store) MarkObjectUploaded(bucket, name, expectedFilename string, siaObject slabs.SealedObject) error {
```
ChrisSchinnerl: I'd use the `content_md5` hash to determine whether to return `ErrObjectModified`. It's not obvious that a filename would give you that

sia/persist/sqlite/objects.go
```diff
@@ -172,16 +173,16 @@ func (s *Store) MarkObjectUploaded(bucket, name string, siaObject slabs.SealedOb
 		res, err := tx.Exec(`
 			UPDATE objects
 			SET object_id = $1, sia_object = $2, filename = NULL, cached_at = $3
-			WHERE bucket_id = $4 AND name = $5 AND object_id IS NULL
```
ChrisSchinnerl, line 175: why loosen the constraint around the file not being uploaded yet?

sia/persist/sqlite/objects.go
```diff
@@ -172,16 +173,16 @@ func (s *Store) MarkObjectUploaded(bucket, name string, siaObject slabs.SealedOb
...
-		`, sqlHash256(objID), sqlSiaObject(siaObject), sqlTime(time.Now()), bid, name)
+			WHERE bucket_id = $4 AND name = $5 AND filename = $6
+		`, sqlHash256(objID), sqlSiaObject(siaObject), sqlTime(time.Now()), bid, name, expectedFilename)
 		if err != nil {
 			return err
 		} else if n, err := res.RowsAffected(); err != nil {
 			return err
 		} else if n == 0 {
-			return fmt.Errorf("object already has an object ID or does not exist")
+			return objects.ErrObjectModified
```
ChrisSchinnerl, line 187: I'd differentiate between the file not existing and the file existing but being modified.

sia/persist/sqlite/objects.go
```diff
@@ -302,6 +303,39 @@ func (s *Store) RemoveOrphanedObject(objectID types.Hash256) error {
...
+			ORDER BY o.size DESC
+		`)
+		if err != nil {
+			return err
+		}
+		defer rows.Close()
+
+		for rows.Next() {
+			var obj objects.PackedObject
+			if err := rows.Scan(&obj.Bucket, &obj.Name, &obj.Filename, &obj.Length, &obj.PartsCount); err != nil {
```
ChrisSchinnerl: The packed object only concerns itself with whether it's multipart or not right? Not the number of parts? So I'd change that field to a boolean and possibly speed up the query with an exists instead of a left join.

sia/packing.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+	return (p.totalSize + p.slabSize - 1) / p.slabSize
+}
+
+func (p *packedObjects) tryAdd(obj objects.PackedObject) bool {
+	newTotal := p.totalSize + obj.Length
+	newSlabs := (newTotal + p.slabSize - 1) / p.slabSize
+
+	// don't exceed the maximum number of slabs per pack
+	if newSlabs > maxSlabsPerPack {
```
ChrisSchinnerl: Why reintroduce this? If we exit early here we won't fill the gaps of the last slab anymore if possible.

sia/packing.go
```diff
@@ -0,0 +1,339 @@
...
+		s.removeUploadQuiet(packObj.Filename)
+		s.logger.Debug("packed object uploaded to Sia",
+			zap.String("bucket", packObj.Bucket),
+			zap.String("name", packObj.Name))
+	}
+
+	return nil
+}
+
+func (s *Sia) removeUploadQuiet(fileName string) {
```
ChrisSchinnerl: This is just called in one spot. You can inline it.

sia/upload.go
ChrisSchinnerl, line 1: This whole file and all the associated methods can now be renamed to just `upload.go`, `uploadLoop`, `ObjectsForUpload` etc. since it's not longer packing specific but just the way we perform uploads in S3d.

ChrisSchinnerl (changes requested):

sia/persist/sqlite/objects.go
peterjan: I can add it but it's redundant because of the constraint, if filename is not null the object id is null. 
I ended up reverting the query anyway to differentiate between modified and deleted

sia/persist/sqlite/objects.go
peterjan: nice catch

sia/packing.go
peterjan: I left you a big comment [here](https://github.com/SiaFoundation/s3d/pull/101#issuecomment-4287460568). I think we should reintroduce it because the results are a lot better with a cap on slabs. The slabs get filled fine with the last step that takes all packs that don't meet the threshold and distributes those objects over existing packs that have room to fit whatever objects are remaining. 

Edit: I updated it again with something I think makes sense. Fetching some numbers now

sia/upload.go
```diff
@@ -0,0 +1,340 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-t.C:
+			s.uploadObjectsIter(ctx)
+		}
+	}
+}
+
+func (s *Sia) uploadObjectsIter(ctx context.Context) {
```
ChrisSchinnerl: (nit) `uploadObjects`

sia/packing.go
ChrisSchinnerl: Can you add a test we can run to reproduce this? My concern is that whatever you ran just happens to work well with whatever setup you used combined with that arbitrary number. 
Some of those numbers to me just can't be explained with a cutoff. e.g. why one would have single object packs while the other one doesn't. They should be just as efficient assuming you upload the same series of objects to both. 

> The slabs get filled fine with the last step that takes all packs that don't meet the threshold and distributes those objects over existing packs that have room to fit whatever objects are remaining.

Are you referring to the second `tryAdd` call? Because that will still return `false` for slabs reaching the `maxSlabsPerPack` limit.

sia/upload.go
peterjan: lint complains about clashing with `UploadObjects` which I expose in testing, do you care a lot?

sia/packing.go
peterjan: @ChrisSchinnerl 

I added a temporary commit that adds `TestTryAddComparison`. It runs the old `TryAdd` vs the new and compares packing 200 objects between 1 and 50mb in size, which is similar to the rclone test I'm running on my node.

```
    upload_test.go:192:                             Old        New
    upload_test.go:193: Packs                        36         30
    upload_test.go:194: Objects packed              199        196
    upload_test.go:195: Total slabs                 123        121
    upload_test.go:196: Max waste                  9.30%      1.76%
    upload_test.go:197: Avg waste                  1.38%      0.46%
    upload_test.go:198: Single-obj packs              4          0
    upload_test.go:199: Max slabs/pack                7          6
```

It's a packing problem where we don't know the size of the inputs beforehand. So obviously this is going to perform better/worse depending on the type of objects users upload and we can't optimize for everything. If you upload 1kb files or all files that already meet the waste threshold, your results are going to be dramatically different.

The max number of slabs is indeed arbitrary, but it's meant to cap the per-thread cycle time of an upload thread to be under 30 seconds. With a connection as slow as 600mbps and 8 workers, that is still the case. With a connection speed of 1gbps every "pack" should be done uploading in 15s which ensures we keep all threads busy.

Can we agree we want to avoid single slab objects, and that we want to have a cap on the maximum slabs per object? If we can agree on that I don't see any issue with the current implementation. I only want to change it because I noticed the previous implementation was sub-optimal.

sia/upload_test.go
```diff
@@ -2,15 +2,218 @@ package sia
...
+
+func TestTryAddComparison(t *testing.T) {
+	const (
+		slabSize       = 40 << 20 // 40 MiB
+		uploadWastePct = 0.10
+		numObjects     = 200
+	)
+
+	// generate a fixed set of random objects with sizes between 1 and 50 MiB
+	rng := frand.NewCustom(make([]byte, 32), 1024, 20)
```
peterjan: note that if you don't seed this you occasionally get similar or better stats with the old implementation due to chance but generally the new implementation performs better

sia/upload.go
ChrisSchinnerl: `performUpload` then? Whatever doesn't trigger the immediate feeling that this is somehow related to a custom iterator.

sia/multipart.go
```diff
@@ -246,14 +246,17 @@ func (s *Sia) UploadPartCopy(ctx context.Context, accessKeyID, srcBucket, srcObj
...
 	if obj.FileName != nil {
-		src, err = s.openUpload(srcBucket, srcObject, obj, opts.Range.Start)
+		src, err = s.openUpload(srcBucket, srcObject, obj.FileName, obj.IsMultipart(), opts.Range.Start)
 		if err != nil {
 			return nil, fmt.Errorf("failed to open source upload: %w", err)
 		}
 	} else {
-		pinnedObj, err := s.sdk.Object(ctx, *obj.ID)
+		if obj.SiaObject == nil {
+			return nil, fmt.Errorf("object metadata not yet synced from indexer")
```
ChrisSchinnerl: (nit) this error doesn't make a lot of sense. We atomically swap the filename for the sia object. There is no delay involved between uploading and syncing metadata.

sia/upload.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+func (p *uploadGroup) slabs() int64 {
+	if p.totalSize == 0 {
+		return 0
+	}
+	return (p.totalSize + p.slabSize - 1) / p.slabSize
+}
+
+func (p *uploadGroup) tryAdd(obj objects.ObjectForUpload) bool {
```
ChrisSchinnerl, line 82: This probably needs an exception in case `p.totalSize` is 0. So that large files that exceed `maxSlabsPerUpload` can be uploaded at all. Otherwise they'll never make it in a group.

ChrisSchinnerl (changes requested):

sia/multipart.go
peterjan: I'll update the message. It's just a sanity check right before we dereference it, it's one of those " // this should never happen" developer errors

sia/upload.go
```diff
@@ -0,0 +1,346 @@
...
+			ready = append(ready, g)
+		} else {
+			remaining = append(remaining, g.objects...)
+		}
+	}
+
+	// try and fill gaps with remaining objects
+	for _, obj := range remaining {
+		for i := range ready {
+			if ready[i].tryAdd(obj) {
```
ChrisSchinnerl: I think you can now remove this loop and the `remaining` slice. Because `tryAdd` will never return `true` here.

Because in the first loop we always try to add all candidates to all groups. So group 1 has seen all candidates, group 2 has seen all that didn't fit in group 1, group 3 the ones that didn't fit in group 1 or 2 and so on.
Also `tryAdd` will never push a group from being above the threshold to below the threshold.

So that means if group `n` didn't make the threshold, group `n+1` also didn't. Because it only had a subset of candidates to work with.

So if you try to `ready[i].tryAdd(obj)` you are executing a `tryAdd` that has already been done before.

Running it against `TestTryAddComparison` without this loop also resulted in the exact same output.

sia/upload.go
```diff
@@ -0,0 +1,346 @@
...
+			if err := s.sdk.DeleteObject(ctx, obj.ID()); err != nil {
+				s.logger.Error("failed to delete pinned object after finalize failure",
+					zap.String("bucket", uploadObj.Bucket),
+					zap.String("name", uploadObj.Name),
+					zap.Error(err))
+			}
+			continue
+		}
+
+		if err := s.removeUpload(uploadObj.Filename); err != nil && !errors.Is(err, os.ErrNotExist) {
```
ChrisSchinnerl: You need to call either `removeUpload` or `removeMultipartUploadDir` depending on whether the upload was a multipart upload. I'd also not ignore `os.ErrNotExist` because it really shouldn't happen unless it was manually deleted.

ChrisSchinnerl (changes requested):

sia/upload.go
peterjan: Hm took me a minute to confirm this. I thought it was possible in the case where you have a large object to begin with, because there we don't accept new objects even it reduces waste. You are right though, once a group is below the waste threshold, the conditions for accepting are tighter than on the first attempt so the second loop never accepts anything.

sia/upload.go
peterjan: Yeah they do exactly the same thing though. I noticed it but I was going to tackle it when we merge in the reference counting delete PR you are working on. I updated `removeMultipartUploadDir` to take an `uploadDir string` so you can call .String() if you have an UploadID and avoid having to parse the UploadID from the filename. 

Should we rename filename to location? That covers the load a bit better now that it can be both a file and a directory

ChrisSchinnerl (dismissed):

sia/objects_test.go
```diff
@@ -685,8 +777,14 @@ func TestObjectMetadataCache(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	} else if obj.FileName == nil {
+		t.Fatal("expected pending upload to have a filename")
+	}
 	sealed := memSDK.SealObject(siaObj)
-	if err := store.MarkObjectUploaded(bucket, object, sealed); err != nil {
+	if err := store.MarkObjectUploaded(bucket, object, obj.ContentMD5, sealed); err != nil {
 		t.Fatal(err)
 	}
```
chris124567, line 790: We can enable TestDeleteObjectUnpin now

ChrisSchinnerl: Good point. @peterjan

sia/persist/sqlite/objects.go
```diff
@@ -415,6 +414,12 @@ func insertOrphan(tx *txn, objectID types.Hash256) error {
 	return err
 }
 
+// removeOrphan deletes objectID from the orphaned_objects table.
+func removeOrphan(tx *txn, objectID types.Hash256) error {
```
ChrisSchinnerl: Do we even need this anymore? It should be impossible for an object to become an orphan and then unorphaned because it gets uploaded again. The SDK would use a new random key every time even if you upload the exact same data again.

ChrisSchinnerl (approved):

chris124567 (approved):

PR #102, opened 2026-03-06
Make sure object gets unpinned from the indexer after deletion in s3d

### LINKED ISSUE #68: Make sure object gets unpinned from the indexer after deletion in s3d
This might sound trivial but it's a bit tricky. Since multiple objects in our database can point to the same object on the indexer, we will need to perform some reference counting. When there are no more objects in the database for a given object id we can delete it but we need to be careful. If another thread adds it back to the database in the meantime we might end up with a local object that is no longer pinned on the indexer.

What we might need here is to wait for the planned breaking SDK changes and https://github.com/SiaFoundation/indexd/pull/671.

Then we can get the object's ID before pinning it and introduce some locking.

chris124567:
Close #68


sia/persist/sqlite/init.sql
```diff
@@ -22,6 +22,7 @@ CREATE TABLE objects (
     cached_at INTEGER NOT NULL,
     PRIMARY KEY (bucket_id, name)
 ) WITHOUT ROWID;
+CREATE INDEX objects_object_id_idx ON objects(object_id);
```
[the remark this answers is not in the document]
chris124567: we aren't doing migrations yet

sia/persist/sqlite/migrations_test.go
```diff
@@ -35,6 +35,7 @@ CREATE TABLE objects (
     cached_at INTEGER NOT NULL,
     PRIMARY KEY (bucket_id, name)
 ) WITHOUT ROWID;
+CREATE INDEX objects_object_id_idx ON objects(object_id);
```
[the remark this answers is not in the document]
chris124567: we aren't doing migrations yet

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/SiaFoundation/s3d
 
-go 1.25
-
-toolchain go1.25.5
+go 1.26.0
```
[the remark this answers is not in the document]
chris124567: we need this due to new indexd

sia/objects.go
```diff
@@ -40,9 +42,15 @@ func (s *Sia) CopyObject(ctx context.Context, accessKeyID, srcBucket, srcObject,
...
-	if err := s.store.PutObject(accessKeyID, dstBucket, dstObject, obj, true); err != nil {
+	orphaned, err := s.store.PutObject(accessKeyID, dstBucket, dstObject, obj, true)
+	if err != nil {
+		s.unpinMu.Unlock()
 		return nil, err
 	}
+	s.unpinMu.Unlock()
+
+	s.tryUnpinObject(ctx, orphaned)
+
```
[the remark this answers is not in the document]
chris124567: We acquire the mutex in tryUnpinObject so I'm sort of confused what is meant by this

sia/sia.go
```diff
@@ -105,6 +110,29 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
+		return
+	}
+	if referenced {
+		return
+	}
+
+	if err := s.sdk.DeleteObject(ctx, *objectID); err != nil && !errors.Is(err, slabs.ErrObjectNotFound) {
+		s.logger.Error("failed to unpin object from indexer", zap.Error(err), zap.Stringer("objectID", objectID))
+	}
+}
```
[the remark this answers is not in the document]
chris124567: Then we'd have to hold the mutex during the SDK call to DeleteObject which I think would slow down things a lot if we had concurrent deletes and writes... not sure there's a good answer to this

sia/objects.go
```diff
@@ -40,9 +42,15 @@ func (s *Sia) CopyObject(ctx context.Context, accessKeyID, srcBucket, srcObject,
 		maps.Copy(obj.Meta, meta)
 	}
 
-	if err := s.store.PutObject(accessKeyID, dstBucket, dstObject, obj, true); err != nil {
+	orphaned, err := s.store.PutObject(accessKeyID, dstBucket, dstObject, obj, true)
```
ChrisSchinnerl: I'd implement a dedicated `CopyObject` method on the store which atomically copies the object in a single transaction. That way you don't need to force atomicity through a mutex. Internally this could be as simple as extracting `GetObject` and `PutObject` into `getObject` and `putObject` helpers which take a transaction and then call them both within a single transaction within `CopyObject` or you implement a custom method if that is easier.

sia/sia.go
```diff
@@ -105,6 +110,29 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
+	s.unpinMu.Unlock()
+	if err != nil {
+		s.logger.Error("failed to check object references", zap.Error(err), zap.Stringer("objectID", objectID))
+		return
+	}
+	if referenced {
+		return
+	}
+
+	if err := s.sdk.DeleteObject(ctx, *objectID); err != nil && !errors.Is(err, slabs.ErrObjectNotFound) {
```
ChrisSchinnerl: The problem with this is that failing to delete the object because the indexer is temporarily unreachable causes you to lose track of orphaned objects.

So what I would do instead is not return orphans from the store methods but atomically insert them into a table for orphaned objects within the same transaction that created them. Then add a background loop to `Sia` which periodically checks for orphans and unpins them from Sia first before it then removes them from the database.

That way we can retry deleting the object if it fails even if we crash or something. We just need to make sure that we ignore the `ErrObjectNotFound` error which you already do.

ChrisSchinnerl (changes requested):

sia/objects_test.go
```diff
@@ -812,3 +814,99 @@ func TestObjectMetadataCache(t *testing.T) {
...
+	_, err = s3Tester.CopyObject(t.Context(), bucket, "A", bucket, "B", types.MetadataDirectiveCopy, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// delete A - should NOT unpin since B still references the same object_id
+	if err := s3Tester.DeleteObject(t.Context(), bucket, "A"); err != nil {
+		t.Fatal(err)
+	}
+	siaBackend.ProcessOrphans(t.Context())
```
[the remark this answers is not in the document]
chris124567: fixed

sia/objects_test.go
```diff
@@ -812,3 +814,99 @@ func TestObjectMetadataCache(t *testing.T) {
...
+	memSDK := NewMemorySDK()
+	log := zaptest.NewLogger(t)
+	dir := t.TempDir()
+	store, err := sqlite.OpenDatabase(filepath.Join(dir, "s3d.sqlite"), log)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { store.Close() })
+
+	siaBackend, err := sia.New(context.Background(), memSDK, store, dir, sia.WithKeyPair(testutil.AccessKeyID, testutil.SecretAccessKey), sia.WithLogger(log))
```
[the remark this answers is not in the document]
chris124567: fixed

sia/sia.go
```diff
@@ -102,9 +108,47 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
+	orphans, err := s.store.OrphanedObjects()
+	if err != nil {
+		s.logger.Error("failed to fetch orphaned objects", zap.Error(err))
+		return
+	}
+	for _, id := range orphans {
+		if err := s.sdk.DeleteObject(ctx, id); err != nil && !errors.Is(err, slabs.ErrObjectNotFound) {
+			s.logger.Error("failed to unpin object from indexer", zap.Error(err), zap.Stringer("objectID", &id))
+			continue
+		}
```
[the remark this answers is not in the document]
chris124567: fixed

sia/persist/sqlite/multipart.go
```diff
@@ -87,6 +89,11 @@ func (s *Store) CompleteMultipartUpload(bucket, name string, uploadID s3.UploadI
 			return fmt.Errorf("found %d parts smaller than minimum size (%d bytes)", smallParts, s3.MinUploadPartSize)
 		}
 
+		oldID, err := previousObjectID(tx, bid, name)
+		if err != nil {
+			return err
+		}
+
```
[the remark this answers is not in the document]
chris124567: fixed

sia/persist/sqlite/objects.go
```diff
@@ -58,87 +63,207 @@ func (s *Store) GetObject(accessKeyID *string, bucket, name string, partNumber *
...
+	`, bid, name, sqlHash256(obj.ID), sqlMD5(obj.ContentMD5),
+		sqlMetaJSON(obj.Meta), obj.Length, sqlTime(time.Now()),
+		sqlSiaObject(obj.SiaObject), sqlTime(obj.CachedAt), updateModTime)
+	if err != nil {
+		return err
+	}
+
+	if oldID != nil && *oldID != obj.ID {
+		return insertOrphan(tx, *oldID)
+	}
```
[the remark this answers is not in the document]
chris124567: fixed

sia/sia.go
```diff
@@ -102,9 +108,51 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
+	for _, id := range orphans {
+		shouldUnpin, err := s.store.RemoveOrphanedObject(id)
+		if err != nil {
+			s.logger.Error("failed to remove orphaned object", zap.Error(err), zap.Stringer("objectID", &id))
+			continue
+		}
+		if !shouldUnpin {
+			continue
+		}
+		if err := s.sdk.DeleteObject(ctx, id); err != nil && !errors.Is(err, slabs.ErrObjectNotFound) {
```
peterjan: This is never retried, so we lose track of the orphan here.

sia/persist/sqlite/objects.go
```diff
@@ -58,85 +63,232 @@ func (s *Store) GetObject(accessKeyID *string, bucket, name string, partNumber *
...
 }
 
-// PutObject stores the given object in the given bucket with the given name or
-// overwrites it if it already exists.  If updatedModTime is true, the
-// `updated_at` time that represents the S3 last modified time will be updated.
-func (s *Store) PutObject(accessKeyID, bucket, name string, obj *objects.Object, updateModTime bool) error {
-	return s.transaction(func(tx *txn) error {
-		bid, err := bucketID(tx, bucket)
+// OrphanedObjects returns all object IDs in the orphaned_objects table.
+func (s *Store) OrphanedObjects() (ids []types.Hash256, err error) {
```
peterjan: This is unbound and potentially a very large list of object IDs. We should pass a limit to process orphans in batches.

sia/persist/sqlite/objects.go
```diff
@@ -58,85 +63,232 @@ func (s *Store) GetObject(accessKeyID *string, bucket, name string, partNumber *
...
-			err := tx.QueryRow(`
-				SELECT object_id, metadata, updated_at, size, content_md5, sia_object, cached_at
-				FROM objects
-				WHERE bucket_id = $1 AND name = $2
-			`, bid, name).Scan((*sqlHash256)(&obj.ID), (*sqlMetaJSON)(&obj.Meta), (*sqlTime)(&obj.LastModified), &obj.Length, (*sqlMD5)(&obj.ContentMD5), &siaObj, (*sqlTime)(&obj.CachedAt))
-			obj.SiaObject = slabs.SealedObject(siaObj)
+// CopyObject atomically reads the source object and writes it to the
+// destination within a single transaction, applying metadata according to the
+// replace flag. Returns the copied object metadata.
+func (s *Store) CopyObject(srcBucket, srcName, dstBucket, dstName string, meta map[string]string, replace bool) (result *objects.Object, err error) {
```
peterjan, line 137: This is great but it's unrelated right? Did you notice and decided to fix it here?

sia/sia.go
```diff
@@ -102,9 +108,51 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
+			return
+		case <-t.C:
+			s.ProcessOrphans(ctx)
+		}
+	}
+}
+
+// ProcessOrphans unpins orphaned objects from the indexer and removes them
+// from the orphaned_objects table.
+func (s *Sia) ProcessOrphans(ctx context.Context) {
```
peterjan, line 138: I would add some more logging in the happy case as well, so we know it's working.

sia/sia.go
```diff
@@ -102,9 +108,51 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
+
 	return sia, nil
 }
 
+// processOrphansLoop periodically processes orphaned objects.
+func (s *Sia) processOrphansLoop(ctx context.Context) {
+	t := time.NewTicker(time.Minute)
+	defer t.Stop()
+
+	// process once immediately at startup
```
peterjan: why? It runs every minute, which seems excessive but probably doesn't really matter

sia/sia.go
peterjan: Only if we processed orphans though, seeing as this runs every minute. We should probably increase that to run every hour or so.

peterjan (changes requested):

sia/persist/sqlite/objects.go
chris124567: Chris asked me to add it: https://github.com/SiaFoundation/s3d/pull/102#discussion_r2918882231

sia/persist/sqlite/objects_test.go
```diff
@@ -709,3 +709,163 @@ func BenchmarkListObjects(b *testing.B) {
...
+	// put first object
+	if err := store.PutObject(accessKeyID, bucket, "a", &objects.Object{
+		ID:         objID,
+		ContentMD5: frand.Entropy128(),
+		Length:     1,
+	}, true); err != nil {
+		t.Fatal(err)
+	}
+
+	// put second object with same ID (simulates CopyObject)
```
ChrisSchinnerl: You can now call `CopyObject` instead.

sia/sia.go
```diff
@@ -102,9 +109,60 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
+			s.ProcessOrphans(ctx)
+		}
+	}
+}
+
+// ProcessOrphans unpins orphaned objects from the indexer and removes them
+// from the orphaned_objects table. It processes up to 100 orphans per call.
+func (s *Sia) ProcessOrphans(ctx context.Context) {
+	const batchSize = 100
+	orphans, err := s.store.OrphanedObjects(batchSize)
```
ChrisSchinnerl: This will now at most prune 100 orphans per hour since it's not inside the loop.

sia/sia.go
```diff
@@ -102,9 +109,60 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
+	if err != nil {
+		s.logger.Error("failed to fetch orphaned objects", zap.Error(err))
+		return
+	}
+	if len(orphans) == 0 {
+		return
+	}
+	var unpinned int
+	for _, id := range orphans {
+		shouldUnpin, err := s.store.OrphanUnreferenced(id)
```
ChrisSchinnerl: Checking whether the orphan is still an orphan shouldn't be necessary here for two reasons. 
First, you process only 100 anyway. It should be quite quick to delete them.

But more importantly, it should be impossible for an orphan to become not orphaned. The only thing that can duplicate object ids is copying objects right? Since that happens atomically, the object will never be considered an orphan.
Once it is an orphan, you can't use `CopyObject` anymore. 

We should double check but I think even the multipart upload api won't let you create more than 1 object with the same object id. Since the upload parts are deleted after completion.

ChrisSchinnerl (changes requested):

sia/sia.go
```diff
@@ -102,9 +108,57 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
+	for {
+		orphans, err := s.store.OrphanedObjects(batchSize)
+		if err != nil {
+			s.logger.Error("failed to fetch orphaned objects", zap.Error(err))
+			return
+		}
+		for _, id := range orphans {
+			if err := s.sdk.DeleteObject(ctx, id); err != nil && !errors.Is(err, slabs.ErrObjectNotFound) {
+				s.logger.Error("failed to unpin object from indexer", zap.Error(err), zap.Stringer("objectID", &id))
+				continue // leave in table for retry
```
peterjan: If we fetch a full batch, but fail to delete all objects, we never break out of the loop and we fetch the same batch over and over again. We should probably exit the loop if we didn't fail to process a single object and wait for the next `ProcessOrphans` tick to avoid spamming the logs.

sia/sia.go
```diff
@@ -102,9 +108,57 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
+		}
+	}
+}
+
+// ProcessOrphans unpins orphaned objects from the indexer and removes them
+// from the orphaned_objects table in batches.
+func (s *Sia) ProcessOrphans(ctx context.Context) {
+	const batchSize = 100
+	var totalUnpinned int
+	for {
```
peterjan, line 141: I would check `ctx` more often. Neither the outer nor inner loop checks it, so we process all orphans (potentially) on shutdown.

sia/sia.go
```diff
@@ -102,9 +108,57 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
+		orphans, err := s.store.OrphanedObjects(batchSize)
```
peterjan, line 148: We have a race condition here right. What happens to an object that was re-uploaded in the mean time, we're effectively unpinning objects from memory from the indexer here. Perhaps we should update `RemoveOrphanedObject` to add something like `AND NOT EXISTS(SELECT 1 FROM objects WHERE object_id = ?))`, then assert a row got removed, which tells us we can safely go ahead and unpin the object from the indexer, if that fails we have to re-add the orphaned object locally though... cc'ing @ChrisSchinnerl for this thoughts on this

peterjan (changes requested):

sia/sia.go
ChrisSchinnerl: Even if you reupload the same object it won't have the same id. Copying should be the only way to create duplicate ids in the database and once an object is orphaned it can't be copied anymore. That's really the underlying guarantee that this hinges on. Might be worthwhile to add a comment here explaining this if this code raises flags without it.

sia/sia.go
ChrisSchinnerl: We should just exit if deleting an object from indexd or the store fails for any reason that is not related to the object not existing.

peterjan (approved):

ChrisSchinnerl (approved):

Issue #50, opened 2025-11-19
Run the compatibility tests against the Sia backend


ChrisSchinnerl:
This will require the following steps:

1. create a new app key on Zeus for the CI
2. manually create an account using that app key
3. store the secret in a GH secret
4. add a new step to the config to run the compatibility tests against the Sia backend
5. fix any bugs that prevent the existing set of enabled tests to fail.

peterjan:
The s3-tests don't reuse objects in their tests, so running all tests using the default redundancy uploads a considerable amount of data and takes a long time (hours). 

We could update the s3-tests repo to change the test structure to re-use objects as much as possible? I don't really like that option because our fork would be harder to maintain. 

We could run the tests using a lower redundancy. I didn't like that at first because I didn't want to pollute Zeus with min-redundancy objects. However, the s3-tests do a teardown step where they remove the test bucket, meaning all objects are deleted immediately anyway.

Are we sure we want to run the entire suit against the Sia backend in the CI at all?

ChrisSchinnerl:
The uploads are pretty tiny right? So we just need to finish upload packing first and then we should be good to go.

peterjan:
> The uploads are pretty tiny right? So we just need to finish upload packing first and then we should be good to go.

Yeah upload packing would be great.

PR #117, opened 2026-04-13
Run S3 tests against Sia

peterjan:
This PR runs S3 tests against a Sia backend. I moved it into a test to leverage our test cluster. Moving it into a test means we can get rid of the tox command.


sia/persist/sqlite/objects.go
```diff
@@ -381,10 +380,6 @@ WHERE o.bucket_id = ?`
 
 		if !result.IsTruncated {
 			result.NextMarker = ""
-		} else if prefix.HasDelimiter && strings.HasSuffix(result.NextMarker, prefix.Delimiter) {
-			// NextMarker is a common prefix. Append \xFF to skip past all objects
-			// with that prefix on the next call.
-			result.NextMarker += "\xFF"
```
peterjan: We don't have to set this on the result because it breaks paging, we set it on re-entry. This broke the s3-tests so figured I would fix it here.

.github/workflows/tox-test.yml
```diff
@@ -11,6 +11,21 @@ jobs:
     permissions:
       contents: read
 
+    services:
+      postgres:
+        image: postgres:17
```
ChrisSchinnerl: I think the latest one is 18?

tox/tox_test.go
```diff
@@ -0,0 +1,311 @@
...
+
+	return listener.Addr().(*net.TCPAddr).Port
+}
+
+// writeS3TestsConf writes an s3tests.conf file to dir with the given port and
+// returns the path to the file.
+func writeS3TestsConf(t *testing.T, dir string, port int) string {
+	t.Helper()
+
+	conf := fmt.Sprintf(`[DEFAULT]
```
ChrisSchinnerl: Why not add this to the `tox` folder directly? When the test gets executed the working directory should be the package directory. So theoretically you should be able to read `./tox.conf` during the test

tox/tox_test.go
ChrisSchinnerl, line 1: This definitely works but if you look at the issue the goal was to run the compatibility tests against an s3d node connected to Zeus: https://github.com/SiaFoundation/s3d/issues/50

To give us a full e2e compatibility testing.

tox/tox_test.go
```diff
@@ -0,0 +1,311 @@
...
+		accessKey: "HIJKLMNOPQRSTUVWXYZA",
+		secretKey: "opqrstuvwxyzabcdefghijklmnopqrstuvwxyzab",
+	}
+	allKeyPairs = []keyPair{mainKeyPair, altKeyPair, tenantKeyPair}
+)
+
+func TestS3(t *testing.T) {
+	log := zap.NewNop()
+
+	t.Run("memory backend", func(t *testing.T) {
```
ChrisSchinnerl: We can drop running the compat tests against the memory backend now imo. Considering that the SDK is now complete and we can test against a Sia cluster we can probably slowly retire it.

sia/persist/sqlite/objects.go
ChrisSchinnerl: Could you break this out into a separate PR with a unit test? Ideally our automated tests catch issues before we run the compat tests.

ChrisSchinnerl (changes requested):

tox/go.mod
ChrisSchinnerl, line 1: This shouldn't require a separate package right?

tox/tox_test.go
peterjan: Let's drop it when we remove the memory backend? I don't care though, it's fun to run tests against things that work and look at all the green checks. Edit: turned out to be annoying so i dropped it.

.github/workflows/tox-test.yml
peterjan: Sure, I'll just remove it so it fetches the latest stable one. Seems like that is what we usually do. I thought I remember we fixed it at 17 for a while because 18 changed something and required a migration but I forgot what that was.

tox/go.mod
peterjan: There's a couple of reasons, like I didn't want `TestS3` to run if you run `go test ./...` in the main module. 

The most important reason though is to control the dependencies. The tox tests require indexd for its testutils, but s3d should not require indexd at all. I thought that was already the case but turns out it wasn't. I swapped out indexd for the siastorage repo seeing it's much smaller and contains everything we need.

edit: hm ok so indexd is an indirect dep anyway so I'll change it and use a build tag

tox/tox_test.go
peterjan: Actually, _technically_, I'm not closing the issue. I'll tackle it next though.

go.mod
```diff
@@ -9,8 +9,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.99.0
 	github.com/mattn/go-sqlite3 v1.14.42
 	go.sia.tech/core v0.19.1
-	go.sia.tech/indexd v0.1.1-0.20260410081129-29ab0c1887a1
-	go.sia.tech/siastorage v0.0.3
```
peterjan: I got rid of `go.sia.tech/siastorage` and created an issue to update `indexd` to latest version to ensure the `sdk` package is no longer available. Right now we use both and it's a little confusing. It also hides the fact that siastorage does not yet export everything you need.

tox/go.mod
peterjan: It's still a separate package but I got rid of the module.

tox/tox_test.go
```diff
@@ -0,0 +1,192 @@
...
+	// })
+}
+
+func runTox(t *testing.T, confPath, testsDir string, args ...string) {
+	t.Helper()
+
+	toxArgs := []string{"run", "--skip-pkg-install", "-c", testsDir, "--"}
+	toxArgs = append(toxArgs, args...)
+
+	cmd := exec.Command("tox", toxArgs...)
```
chris124567: Use CommandContext with t.Context()

tox/tox_test.go
```diff
@@ -0,0 +1,192 @@
...
+	var buf bytes.Buffer
+	cmd.Stdout = &buf
+	cmd.Stderr = &buf
+
+	err := cmd.Run()
+	output := buf.String()
+	if err != nil {
+		t.Fatalf("tox failed: %v\n%s", err, output)
+	} else if bytes.Contains(buf.Bytes(), []byte("FAILED")) {
+		t.Fatalf("tox reported failures:\n%s", output)
```
chris124567: If tox reports failures does it still exit with code 0? Bc `err` will not equal `nil` if it exits with an error code

peterjan: I added the `else if` as a defensive check and considered it harmless. I think I ran into it when my tests were timing out. I figured it's harmless to keep but I think it's dead code, I'll remove it.

chris124567 (dismissed):

chris124567 (dismissed):

ChrisSchinnerl (dismissed):

ChrisSchinnerl:
@peterjan something is still failing

peterjan:
> @peterjan something is still failing

dammit, last minute addressing of a code review, put `min` instead of `max`

ChrisSchinnerl (approved):

PR #125, opened 2026-04-16
Fix ListObjects 

peterjan:
This PR removes `\xFF` from the `NextMarker`, it break pagination. We don't need to set it because it'll get set on re-entry. Added a regression test to confirm.


chris124567 (approved):
LGTM

At the time I added this, we didn't have `adjustMarkerForCommonPrefix` which I think is why it was necessary

ChrisSchinnerl (approved):

PR #129, opened 2026-04-20
Refactor sia backend to always store uploads on disk first

ChrisSchinnerl:
This PR updates s3d to always store uploads on disk first. A background loop will be added in a follow up which offloads objects to the Sia network. 

Regular uploads end up as a single file in the `./uploads` directory while multipart uploads are stored in a folder matching the upload's id. As long as files aren't uploaded yet they will be served from disk. 

This allows us to streamline the upload code by having all uploads go through the same process. Also leading to increased packing efficiency since we can fill the padding of a multipart or regular object upload with smaller objects.


sia/objects/multipartreader.go
```diff
@@ -25,18 +25,37 @@ type MultipartReader struct {
...
 }
 
-// NewReader creates a new Reader for the given multipart upload.
-func NewReader(partsDir string, parts []Part) (*MultipartReader, error) {
+// NewReader creates a new Reader for the given multipart upload starting at
+// the given byte offset. Parts that fall entirely before the offset are
+// skipped, and within the first relevant part the file is seeked past the
+// remaining bytes.
+func NewReader(partsDir string, parts []Part, offset int64) (*MultipartReader, error) {
 	for _, part := range parts {
```
peterjan, line 33: `nit` this is awkward now that we introduced an offset

sia/objects/multipartreader.go
```diff
@@ -25,18 +25,37 @@ type MultipartReader struct {
...
+	}
+
+	r := &MultipartReader{
 		remainingParts: parts,
 		partsDir:       partsDir,
-	}, nil
+	}
+
+	// discard remaining bytes within the first part so the hasher stays correct
+	if offset > 0 && len(parts) > 0 {
```
peterjan, line 52: `nit` the `len(parts)` check is a little odd? Shouldn't we fail on whatever the error is if you try and discard the first offset bytes on an empty reader?

sia/persist/sqlite/init.sql
```diff
@@ -13,13 +13,16 @@ CREATE TABLE buckets (
...
     content_md5 BLOB NOT NULL,
     metadata TEXT NOT NULL,
     size INTEGER NOT NULL,
     updated_at INTEGER NOT NULL,
-    sia_object BLOB NOT NULL,
+    filename TEXT, -- name of file for regular uploads or dir for multipart uploads
+    sia_object BLOB,
     cached_at INTEGER NOT NULL,
+    -- file is either stored on disk, on Sia or empty.
+    CHECK ((sia_object IS NULL AND object_id IS NULL AND filename IS NOT NULL) OR (sia_object IS NOT NULL AND object_id IS NOT NULL AND filename IS NULL) OR size = 0),
```
peterjan: If we are going to explicit about it, we should check for `OR (object_id IS NULL AND filename IS NULL AND size = 0)` and not just size is 0 because that would allow empty states. I actually updated this in my upload PR after a test highlighted this issue.

sia/objects.go
```diff
@@ -144,8 +187,36 @@ func (s *Sia) headOrGetObject(ctx context.Context, accessKeyID *string, bucket,
...
+			if err != nil {
+				return nil, err
+			} else if obj.SiaObject == nil {
+				return nil, fmt.Errorf("pending upload file disappeared and object is not on Sia")
+			}
+		} else if err != nil {
+			return nil, fmt.Errorf("failed to open pending upload file: %w", err)
+		} else {
+			if resp.Range != nil {
+				resp.Body = io.NopCloser(io.LimitReader(rc, resp.Range.Length))
```
peterjan: can we use a `io.NopCloser` here? won't this leak file descriptors for every range req? We need to delegate Close to rc properly

sia/multipart.go
```diff
@@ -94,7 +136,7 @@ func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object string
 	}
 
 	// create part directory
-	partDir := filepath.Join(s.directory, uploadID.String(), fmt.Sprintf("%d", opts.PartNumber))
+	partDir, err := s.ensureMultipartPartDir(uploadID, opts.PartNumber)
```
peterjan, line 134: you don't check `err`

peterjan (changes requested):

sia/objects/multipartreader.go
ChrisSchinnerl: It's not amazing but cleans things up a bit while not requiring a full implementation of a seeker. The latter would require keeping all parts in memory and an offset and specifying the offset when we open the reader means we get the best of both worlds. But I'm open to suggestions/diffs with better solutions.

sia/objects/multipartreader.go
ChrisSchinnerl: I wanted this to behave the same way as if we didn't have an `offset` and instead discarded the bytes after calling `NewReader`. But without actually doing that so that we avoid opening a lot of file handles for no reason.

peterjan (dismissed):

sia/objects/multipartreader.go
peterjan: I mean if the offset is still greater than zero and we don't have any parts, shouldn't it error out.

sia/multipart.go
```diff
@@ -206,15 +248,32 @@ func (s *Sia) UploadPartCopy(ctx context.Context, accessKeyID, srcBucket, srcObj
 		return nil, s3errs.ErrEntityTooLarge
 	}
 
-	// fetch pinned object from the indexer
-	pinnedObj, err := s.sdk.Object(ctx, obj.ID)
-	if err != nil {
-		return nil, fmt.Errorf("failed to fetch object from indexer: %w", err)
+	// open a reader for the requested range of the source object
+	var src io.ReadCloser
+	if obj.FileName != nil {
```
chris124567, line 248: In `getObject` in the store we don't set FileName on the returned object so I don't think this will ever be triggered

sia/multipart.go
```diff
@@ -336,24 +388,18 @@ func (s *Sia) CompleteMultipartUpload(ctx context.Context, accessKeyID, bucket,
...
 	// assert the upload directory exists
-	uploadDir := filepath.Join(s.directory, uploadID.String())
+	uploadDir := s.multipartUploadDir(uploadID)
 	if _, err := os.Stat(uploadDir); err != nil {
 		return nil, fmt.Errorf("failed to stat upload directory: %w", err)
 	}
 
 	// create reader for the completed object
-	r, err := objects.NewReader(uploadDir, completed)
+	r, err := objects.NewReader(uploadDir, completed, 0)
```
chris124567: We can just remove this right

sia/objects/multipartreader.go
ChrisSchinnerl: This is the same behavior as `Seek` has. You can seek past the end of a file just fine. The first Read then returns EOF.

sia/multipart.go
ChrisSchinnerl: We should:
```
		var objectID sql.Null[sqlHash256]
		var siaObj sql.Null[sqlSiaObject]
		err := tx.QueryRow(`
			SELECT filename, object_id, metadata, updated_at, size, content_md5, sia_object, cached_at
			FROM objects
		`, bid, name).Scan(&obj.FileName, &objectID, (*sqlMetaJSON)(&obj.Meta), (*sqlTime)(&obj.LastModified), &obj.Length, (*sqlMD5)(&obj.ContentMD5), &siaObj, (*sqlTime)(&obj.CachedAt))
```

First argument to `Scan`.

peterjan (approved):

sia/objects.go
```diff
@@ -144,8 +192,36 @@ func (s *Sia) headOrGetObject(ctx context.Context, accessKeyID *string, bucket,
...
-	// refresh cached object metadata if needed
-	siaObj, err := s.refreshSiaObject(ctx, *accessKeyID, bucket, object, obj)
+	// read from disk if the object hasn't been uploaded yet
+	if obj.FileName != nil {
+		var offset int64
+		if resp.Range != nil {
+			offset = resp.Range.Start
+		}
+		rc, err := s.openUpload(bucket, object, obj, offset)
+		if os.IsNotExist(err) {
```
peterjan: I think this should be `errors.Is(err, os.ErrNotExist)` because openUpload might wrap the stats error

sia/objects.go
```diff
@@ -239,33 +317,44 @@ func (s *Sia) PutObject(ctx context.Context, accessKeyID string, bucket, object
...
 			return nil, fmt.Errorf("failed to read object data: %w", err)
 		}
 	} else {
-		// upload the data
-		obj, err := s.sdk.Upload(ctx, lr)
+		// save the object
+		randFileName := randObjectName(bucket, object)
+		objPath = filepath.Join(s.uploadDir(), randFileName)
+		fileName = &randFileName
+		f, err := os.Create(objPath)
```
peterjan, line 335: Is there a chance this exceeds the 255 bytes (the per-component limit on ext4/APFS/NTFS) for keys longer than ~170 chars? The names returned by randObjectName are quite long and we might run into `ENAMETOOLONG` errors here.

peterjan (changes requested):

peterjan (approved):

PR #134, opened 2026-04-21
Run tests on all platforms

### LINKED ISSUE #130: Run tests for windows and Mac OS as well
Run the tests in a matrix like we do for other software (e.g. hostd)

peterjan:
@ChrisSchinnerl can you update the repo rules or branch protection rules to check for `test / success`. Right now it's checking for `TestExpected — Waiting for status to be reported` which never completes.

Closes https://github.com/SiaFoundation/s3d/issues/130


sia/upload_test.go
```diff
@@ -15,6 +16,10 @@ import (
 // still open, and the open handle can still be read from and matches the
 // original data.
 func TestOpenAndRemoveUpload(t *testing.T) {
+	if runtime.GOOS == "windows" {
```
peterjan: Temp skip, still figuring this out. Considering that we have an issue removing multipart uploads I'm wondering whether we shouldn't change strategy for deleting open files all together and use lock files maybe?

sia/persist/sqlite/objects_test.go
```diff
@@ -579,6 +582,7 @@ func BenchmarkListObjects(b *testing.B) {
 	if err != nil {
 		b.Fatal(err)
 	}
+	b.Cleanup(func() { store.Close() })
```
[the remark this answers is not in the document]
peterjan: we don't really care about the error, calling cleanup with store.Close is consistent with all other locations

peterjan:
Moved it to `DRAFT` until `chris/upload-locking` gets merged. That simplifies this PR greatly since we don't have to deal with a special way to open a file since we defer deletion.

sia/multipart.go
```diff
@@ -410,3 +407,15 @@ func (s *Sia) CompleteMultipartUpload(ctx context.Context, accessKeyID, bucket,
...
+
+func syncDir(path string) error {
+	// windows does not support fsync on directories
+	if runtime.GOOS == "windows" {
+		return nil
+	}
+	dir, err := os.Open(path)
+	if err != nil {
+		return err
+	}
```
ChrisSchinnerl, line 415: It might make sense to move the windows check down here after opening the dir. So that we at least know the dir exists. The call site handles `os.ErrNotExist` but that error will never be returned for windows right now. Leading to inconsistencies across operating systems.

ChrisSchinnerl (changes requested):

sia/multipart.go
peterjan: Dammit, nice catch.

ChrisSchinnerl (approved):

PR #136, opened 2026-04-23
Defer deletion of opened uploads until the file isn't used anymore

ChrisSchinnerl:
S3d serves files that aren't uploaded yet from disk. While doing so, the file might get uploaded and end up being deleted.
Depending on the operating system and file system this can lead to all sorts of different behavior.

This PR solves this by adding some basic reference counting which defers the deletion of the file until nothing is using it anymore.

### DIFF
--- a/sia/objects.go
+++ b/sia/objects.go
@@ -63,7 +63,11 @@ func (s *Sia) lockUpload(filename string) func() {
 		defer s.lockedUploadsMu.Unlock()
 
 		lu.refCount--
-		if lu.refCount == 0 {
+		if lu.refCount <= 0 {
+			_, locked := s.lockedUploads[filename]
+			if !locked {
+				panic(fmt.Sprintf("unlock called for filename %s that is not locked", filename))
+			}
 			delete(s.lockedUploads, filename)
 			if lu.deleted {
 				if err := os.RemoveAll(filepath.Join(s.uploadDir(), filename)); err != nil {
@@ -73,16 +77,19 @@ func (s *Sia) lockUpload(filename string) func() {
 				}
 			}
 		}
-		return
 	}
 }
 
-func (s *Sia) openUpload(bucket, name string, filename *string, multipart bool, r *s3.ObjectRange) (io.ReadCloser, error) {
+func (s *Sia) openUpload(bucket, name string, filename *string, multipart bool, r *s3.ObjectRange) (_ io.ReadCloser, err error) {
 	if filename == nil {
 		return nil, os.ErrNotExist
 	}
 	unlock := s.lockUpload(*filename)
-	defer unlock()
+	defer func() {
+		if err != nil {
+			unlock()
+		}
+	}()
 
 	var offset int64
 	if r != nil {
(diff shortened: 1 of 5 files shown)


sia/objects.go
```diff
@@ -24,46 +24,113 @@ import (
...
+		return
+	}
+}
+
+func (s *Sia) openUpload(bucket, name string, filename *string, multipart bool, r *s3.ObjectRange) (io.ReadCloser, error) {
 	if filename == nil {
 		return nil, os.ErrNotExist
 	}
+	unlock := s.lockUpload(*filename)
+	defer unlock()
```
peterjan: wait a minute... 👀 this completely defeats the purpose right? Tests should be updated to catch this

sia/objects.go
```diff
@@ -24,46 +24,113 @@ import (
...
+						zap.String("filename", filename),
+						zap.Error(err))
+				}
+			}
+		}
+		return
+	}
+}
+
+func (s *Sia) openUpload(bucket, name string, filename *string, multipart bool, r *s3.ObjectRange) (io.ReadCloser, error) {
```
peterjan: could add a test that covers `r *s3.ObjectRange` - I believe it's untested at the moment

peterjan (changes requested):

sia/objects.go
```diff
@@ -24,46 +24,113 @@ import (
...
+			delete(s.lockedUploads, filename)
+			if lu.deleted {
+				if err := os.RemoveAll(filepath.Join(s.uploadDir(), filename)); err != nil {
+					s.logger.Error("failed to remove upload upon unlock",
+						zap.String("filename", filename),
+						zap.Error(err))
+				}
+			}
+		}
+		return
```
peterjan: The return function doesn't have return variables, so this return is redundant.

sia/objects.go
ChrisSchinnerl: `TestGetAndHeadObject` has a couple of subtests that set it.

sia/objects.go
```diff
@@ -24,46 +24,120 @@ import (
...
+		io.Reader
+		c      io.Closer
+		unlock func()
+	}
+)
+
+func (lr *lockedUploadReader) Close() error {
+	err := lr.c.Close()
+	lr.unlock()
+	return err
```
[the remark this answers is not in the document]
ChrisSchinnerl: I don't think any path actually does that.

sia/objects.go
```diff
@@ -24,46 +24,120 @@ import (
...
+			}
+			delete(s.lockedUploads, filename)
+			if lu.deleted {
+				if err := os.RemoveAll(filepath.Join(s.uploadDir(), filename)); err != nil {
+					s.logger.Error("failed to remove upload upon unlock",
+						zap.String("filename", filename),
+						zap.Error(err))
+				}
+			}
+		}
```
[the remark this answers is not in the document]
ChrisSchinnerl: It's only 1 operation per finished upload in a background goroutine. For which it should be fast enough. Removal can't happen outside the lock without adding a new race.

peterjan (approved):

PR #137, opened 2026-04-23
Sync object metadata

### LINKED ISSUE #133: Improved Sync
S3d currently remembers the last time it fetched the metadata from the indexer to be able to periodically refresh the cache. That's because the object listing endpoint wasn't aware of migrations in the past.

Now that it is, we can change this approach to reflect what Sia Mobile does which is periodically checking for changes and then syncing those changes to keep the metadata up-to-date without periodically refetching it. 

That simplifies the download logic since we no longer need any fallback mechanism. Instead we can assume the metadata we have is the most recent one.

peterjan:
Description added by @chris124567:
Close https://github.com/SiaFoundation/s3d/issues/133


sia/sia.go
```diff
@@ -196,3 +199,80 @@ func (s *Sia) LoadSecret(ctx context.Context, accessKeyID string) (auth.SecretAc
 	}
 	return slices.Clone(secret), nil
 }
+
+// syncMetadataLoop periodically syncs object metadata from the indexer.
+func (s *Sia) syncMetadataLoop(ctx context.Context) {
+	t := time.NewTicker(5 * time.Minute)
```
chris124567: Should this be lower?  cc @ChrisSchinnerl

sia/persist/sqlite/migrations_test.go
```diff
@@ -75,13 +74,15 @@ CREATE TABLE object_parts (
...
+    sia_object_id BLOB PRIMARY KEY
 );
 
 CREATE TABLE global_settings (
 	id INTEGER PRIMARY KEY NOT NULL DEFAULT 0 CHECK (id = 0), -- enforce a single row
 	db_version INTEGER NOT NULL, -- used for migrations
-	app_key BLOB
+	app_key BLOB,
+	objects_cursor_at INTEGER NOT NULL DEFAULT 0,
+	objects_cursor_key BLOB NOT NULL DEFAULT X'0000000000000000000000000000000000000000000000000000000000000000'
```
ChrisSchinnerl: (nit) these names are not quite obvious. Maybe `last_sync_at` and `last_sync_key` would convey the purpose a bit better.

sia/sia.go
ChrisSchinnerl: I'd probably make it a lot higher honestly. We don't need the same instant sync across devices that Sia Storage does. So I think we should be able to get away with an immediate sync on startup + a sync every 24 hours.

sia/sia.go
```diff
@@ -196,3 +199,80 @@ func (s *Sia) LoadSecret(ctx context.Context, accessKeyID string) (auth.SecretAc
...
+	defer t.Stop()
+
+	// sync once on startup
+	s.syncMetadataIter(ctx)
```
ChrisSchinnerl: Same comment as on the last PR. Satisfying the "confusing-naming" linter by introducing confusing naming seems like the opposite of what it's for. 

It should be fine to disable the linter for the `export_test.go` files and then just call it `syncMetadata`. The exported test methods not matching the unexpected prod methods seems unnecessarily confusing too.

ChrisSchinnerl (changes requested):

sia/persist/sqlite/migrations_test.go
peterjan: I went from `last_sync_at` to `objects_cursor_at` to make it more obvious and to convey the purpose better 😂. I'll change it back though, imo "sync" is very generic.

sia/sia.go
```diff
@@ -196,3 +199,82 @@ func (s *Sia) LoadSecret(ctx context.Context, accessKeyID string) (auth.SecretAc
...
+			s.logger.Error("failed to fetch object events", zap.Error(err))
+			break
+		} else if len(events) == 0 {
+			break
+		}
+
+		var failed bool
+		for _, ev := range events {
+			if ev.Deleted || ev.Object == nil {
+				continue
```
ChrisSchinnerl, line 316: (nit) maybe log a debug line here with the object id of the object that got deleted. It will be nice to see that an object got deleted in DeleteObject and then have the confirmation logged when syncing the deletion from Sia.

ChrisSchinnerl (dismissed):

chris124567 (commented):
LGTM but there's a boatload of conflicts now due to upload packing 😅

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	go.sia.tech/core v0.19.1
 	go.sia.tech/coreutils v0.21.2
 	go.sia.tech/indexd v0.1.1-0.20260410081129-29ab0c1887a1
-	go.sia.tech/siastorage v0.0.4-0.20260410121042-fa1e8e354ff0
+	go.sia.tech/siastorage v0.0.4-0.20260422085341-e099028d6370
```
peterjan: Merge blocker, this PR uses https://github.com/SiaFoundation/sia-storage-go/pull/8

ChrisSchinnerl (dismissed):

go.mod
ChrisSchinnerl: @peterjan that's merged now. You can update the commit you target to master.

peterjan: `main`, unfortunately, can we change that?

sia/sia.go
```diff
@@ -262,3 +264,84 @@ func (s *Sia) LoadSecret(ctx context.Context, accessKeyID string) (auth.SecretAc
...
+			if ev.Deleted {
+				s.logger.Debug("skipping deleted object event", zap.Stringer("objectID", &ev.Key))
+				continue
+			} else if ev.Object == nil {
+				s.logger.Warn("skipping event with nil object", zap.Stringer("objectID", &ev.Key))
+				continue
+			}
+
+			sealed := s.sdk.SealObject(*ev.Object)
+			updated, err := s.store.UpdateSiaObject(objects.SiaObject{ID: sealed.ID(), Sealed: sealed})
```
chris124567: Should we do do a batch update here instead of one object at a time?  If it's only running every 24 hours there could be a large number of updates

sia/objects.go
```diff
@@ -309,45 +311,6 @@ func (s *Sia) headOrGetObject(ctx context.Context, accessKeyID *string, bucket,
...
-// refreshSiaObject refreshes the object's cached Sia object if it is missing
-// or stale. Returns the unsealed sdk.Object for use in downloads.
-func (s *Sia) refreshSiaObject(ctx context.Context, obj *objects.Object) (siaObj sdk.Object, err error) {
-	if obj.ID == nil || obj.SiaObject == nil {
-		return sdk.Object{}, fmt.Errorf("object hasn't been uploaded yet") // should never happen
-	}
-	cached := !obj.CachedAt.IsZero()
-
-	// if cache is fresh, unseal and return
-	cachedUntil := obj.CachedAt.Add(metadataCacheLifetime)
```
chris124567, line 322: We can remove metadataCacheLifetime now

sia/sia.go
peterjan: Not sure if that's really necessary but I added it anyway.

chris124567 (approved):

chris124567 (changes requested):
fix conflicts otherwise LGTM

peterjan:
> fix conflicts otherwise LGTM

merge hell 😅

chris124567 (approved):

ChrisSchinnerl (approved):

PR #139, opened 2026-04-24
Add configuration option to limit size of on-disk files

### LINKED ISSUE #138: Add configuration option to limit size of on-disk files
When using a tool like rclone locally it's quite likely that files get stored and packed on disk faster than they can get offloaded to Sia.
We should consider a mechanism that allows us to block uploads once a certain amount of data is reached on disk. e.g. 10GB or so by default.

chris124567:
Close #138


cmd/s3d/main.go
```diff
@@ -213,6 +213,9 @@ func main() {
 		siaOpts = append(siaOpts, sia.WithKeyPair(kp.AccessKey, kp.SecretKey))
 	}
 	siaOpts = append(siaOpts, sia.WithLogger(log.Named("backend")))
+	if cfg.Sia.DiskUsageLimit > 0 {
+		siaOpts = append(siaOpts, sia.WithDiskUsageLimit(cfg.Sia.DiskUsageLimit))
+	}
```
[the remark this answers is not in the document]
chris124567: fixed

sia/objects.go
```diff
@@ -72,6 +86,16 @@ func (s *Sia) removeUpload(fileName string) error {
...
 func (s *Sia) CopyObject(ctx context.Context, accessKeyID, srcBucket, srcObject, dstBucket, dstObject string, replace bool, meta map[string]string) (*s3.CopyObjectResult, error) {
+	src, err := s.store.GetObject(&accessKeyID, srcBucket, srcObject, nil)
+	if err != nil {
+		return nil, err
+	}
+	if src.FileName != nil {
+		if err := s.checkDiskUsage(src.Length); err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
chris124567: fixed

sia/objects.go
```diff
@@ -307,6 +331,10 @@ func (s *Sia) PutObject(ctx context.Context, accessKeyID string, bucket, object
 		return nil, err
 	}
 
+	if err := s.checkDiskUsage(opts.ContentLength); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
chris124567: fixed

sia/persist/sqlite/objects.go
```diff
@@ -14,6 +14,25 @@ import (
...
+	var usage uint64
+	err := s.transaction(func(tx *txn) error {
+		var objectsSize, partsSize uint64
+		if err := tx.QueryRow(`SELECT COALESCE(SUM(size), 0) FROM objects WHERE filename IS NOT NULL`).Scan(&objectsSize); err != nil {
+			return err
+		}
+		if err := tx.QueryRow(`SELECT COALESCE(SUM(content_length), 0) FROM multipart_parts`).Scan(&partsSize); err != nil {
+			return err
+		}
+		usage = objectsSize + partsSize
```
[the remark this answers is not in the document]
chris124567: fixed

sia/persist/sqlite/objects.go
```diff
@@ -14,6 +14,25 @@ import (
...
+		return nil
+	})
+	return usage, err
```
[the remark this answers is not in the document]
chris124567: Added a cache... we could have a counter but it'd be a lot more code

cmd/s3d/main.go
```diff
@@ -47,7 +47,8 @@ var cfg = Config{
 		},
 	},
 	Sia: Sia{
-		IndexerURL: "https://sia.storage",
+		IndexerURL:     "https://sia.storage",
+		DiskUsageLimit: 10 * (1 << 30), // 10 GiB
```
peterjan, line 46: Feels low to me but might be fine.

cmd/s3d/main.go
```diff
@@ -213,6 +214,9 @@ func main() {
 		siaOpts = append(siaOpts, sia.WithKeyPair(kp.AccessKey, kp.SecretKey))
 	}
 	siaOpts = append(siaOpts, sia.WithLogger(log.Named("backend")))
+	if cfg.Sia.DiskUsageLimit > 0 {
```
peterjan: It's a uint64 and it's disabled if you set it to 0 so you don't need this check

sia/sia.go
```diff
@@ -75,6 +86,11 @@ type Sia struct {
 	slabSize       int64
 	uploadWastePct float64
 	uploadDisabled bool
+	diskUsageLimit uint64
+
+	diskUsageMu      sync.Mutex
+	diskUsageCached  uint64
```
peterjan: The cache is not a good idea because the limit should be a hard limit. If we cache this value you can get in scenarios where the limit is exceeded, even if it's for a short amount of time. I think a better approach would be to centralise all disk operations, query the database on startup and then track the disk usage in memory. Wouldn't that solve most of our problems?

sia/objects.go
```diff
@@ -24,6 +24,45 @@ import (
...
+	usage, err := s.store.DiskUsage()
+	if err != nil {
+		return 0, err
+	}
+	s.diskUsageCached = usage
+	s.diskUsageUpdated = time.Now()
+	return usage, nil
+}
+
+func (s *Sia) checkDiskUsage(size int64) error {
```
peterjan: `diskUsage` and `checkDiskUsage` have too much in common, I'd refactor it so one calls the other or get rid of one entirely

peterjan (changes requested):

sia/sia.go
peterjan: Probably should've tagged @ChrisSchinnerl on this one considering it's a rewrite and we might not want to keep track of this in memory. I don't think it's that big of a deal though and it has the benefit of having to centralise disk operations, which is nice. It can maybe even expose bugs in the future if we notice our in memory value differs greatly from the database value.

sia/objects.go
```diff
@@ -24,6 +24,45 @@ import (
...
+	if time.Since(s.diskUsageUpdated) >= time.Minute {
+		usage, err := s.store.DiskUsage()
+		if err != nil {
+			return fmt.Errorf("failed to check disk usage: %w", err)
+		}
+		s.diskUsageCached = usage
+		s.diskUsageUpdated = time.Now()
+	}
+	if s.diskUsageCached+uint64(size) > s.diskUsageLimit {
+		return s3errs.ErrStorageLimitExceeded
```
ChrisSchinnerl: One thing that might not have been obvious from the issue is that once you reach the limit, uploads shouldn't fail but block. 
So if you upload tons of 1MiB files, your upload speed will match your disk speed but once you hit the limit it will slow down and match the speed of the background loop uploading to Sia. 

I'd also treat the limit as a soft limit. e.g. if you have 9 objects incoming with 1GB each and 1 5GB object at a 10GB limit, increment the limit to 9GB, then add the 5GB object which makes it 14GB and from that moment on wait until we fall back to <10GB. 

Similar exception for ongoing multipart uploads. If a multipart upload consists of 3 parts of 5GB each we can't stop at 10GB and then wait forever. We need to recognize that the upload is ongoing and then just add the third part which makes it 15GB total.

Apart from that though I'd stick to what PJ already said. Figure out the size on startup and from then on track it in memory.

ChrisSchinnerl (changes requested):

sia/objects.go
peterjan: Aha, that makes more sense. I thought you mentioned uploads should be prevented and went back to the issue to check and it said "that allows us to block uploads", which I interpreted as "prevent uploads". Perhaps Christopher interpreted that the same way I did. Adds quite a bit of complexity though, especially with multipart uploads 😅 .

cmd/s3d/main.go
chris124567: fixed

sia/sia.go
```diff
@@ -150,6 +166,11 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
 	if err := os.MkdirAll(dir, 0700); err != nil {
 		return nil, fmt.Errorf("failed to create directory %q: %w", dir, err)
 	}
+	diskUsage, err := sia.store.DiskUsage()
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine disk usage: %w", err)
+	}
+	sia.diskUsage = diskUsage
```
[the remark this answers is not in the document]
chris124567: it's not a super expensive query

sia/persist/sqlite/init.sql
```diff
@@ -26,6 +26,7 @@ CREATE TABLE objects (
     PRIMARY KEY (bucket_id, name)
 ) WITHOUT ROWID;
 CREATE INDEX objects_object_id_idx ON objects(object_id);
+CREATE INDEX objects_pending_idx ON objects(filename) WHERE filename IS NOT NULL;
```
peterjan: `nit` even though I like descriptive index names over column names, we should probably rename it to `objects_filename_idx`

sia/persist/sqlite/multipart.go
```diff
@@ -38,9 +38,11 @@ func (s *Store) CreateMultipartUpload(bucket, name string, uploadID s3.UploadID,
 // CompleteMultipartUpload finalizes a multipart upload by creating the object
 // and transferring parts from the upload to the object. If the overwritten
 // object's ID has no remaining references, it is inserted into the
-// orphaned_objects table.
-func (s *Store) CompleteMultipartUpload(bucket, name string, uploadID s3.UploadID, contentMD5 [16]byte, contentLength int64) error {
-	return s.transaction(func(tx *txn) error {
+// orphaned_objects table. If the overwrite leaves a previously pending file
+// unreferenced, its filename is returned so the caller can remove it from disk.
+func (s *Store) CompleteMultipartUpload(bucket, name string, uploadID s3.UploadID, contentMD5 [16]byte, contentLength int64) (*string, error) {
```
peterjan: Nice find! Consider extract this into a separate PR though, it's a decent amount of lines, I'd say it's half the PR so it would make for a nice PR on its own. Considering this wasn't caught in tests I would make sure we have a test that verifies the orphaned file is removed both on CompleteMultipart, CopyObject and PutObject.

peterjan (dismissed):
PR looks great now, I would extract the orphaned file fix into a separate PR and extend testing there so this doesn't go unnoticed. Nice find

peterjan (commented):
Going to wait until https://github.com/SiaFoundation/s3d/pull/151 is in to get a super clean diff, PR looks good though 👍

sia/sia.go
```diff
@@ -102,6 +116,7 @@ type Store interface {
 	CreateBucket(accessKeyID, bucket string) error
 	DeleteBucket(accessKeyID, bucket string) error
 	DeleteObject(accessKeyID, bucket string, objectID s3.ObjectID) (*string, error)
+	DiskUsage() (uint64, error)
 	GetObject(accessKeyID *string, bucket, object string, partNumber *int32) (*objects.Object, error)
```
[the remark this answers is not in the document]
chris124567: not an issue atm

sia/objects.go
```diff
@@ -343,6 +447,24 @@ func (s *Sia) PutObject(ctx context.Context, accessKeyID string, bucket, object
...
+	var objPath string
+	defer func() {
+		if err == nil {
+			return
+		}
+		if objPath != "" {
+			if err := os.Remove(objPath); err != nil && !errors.Is(err, os.ErrNotExist) {
+				s.logger.Error("failed to remove pending upload file after error",
+					zap.String("path", objPath),
+					zap.Error(err))
```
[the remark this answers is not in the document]
chris124567: fixed

sia/multipart.go
```diff
@@ -234,6 +251,20 @@ func (s *Sia) UploadPartCopy(ctx context.Context, accessKeyID, srcBucket, srcObj
...
+	}
+	if err := s.addDiskUsage(ctx, opts.Range.Length, allowExcess); err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err != nil {
+			s.releaseDiskUsage(opts.Range.Length)
+		}
+	}()
+
```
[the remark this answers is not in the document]
chris124567: fixed

sia/multipart.go
```diff
@@ -119,10 +119,24 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
...
+		return s.store.HasMultipartUpload(bucket, object, uploadID)
+	}
+	if err := s.addDiskUsage(ctx, opts.ContentLength, allowExcess); err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err != nil {
+			s.releaseDiskUsage(opts.ContentLength)
+		}
+	}()
```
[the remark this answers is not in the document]
chris124567: fixed

sia/multipart.go
```diff
@@ -107,7 +107,7 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
 	}
 
 	// remove multipart upload directory
-	if err := s.removeUpload(uploadID.String()); err != nil && !errors.Is(err, os.ErrNotExist) {
+	if err := s.removeAndRelease(s.multipartUploadPath(uploadID.String())); err != nil && !errors.Is(err, os.ErrNotExist) {
```
ChrisSchinnerl: This feels like it should be `removeUpload` like before? To avoid reintroducing the issue of deleting a multipart upload that is currently in use. That should call `removeAndRelease` implicitly anyway.

sia/objects.go
```diff
@@ -20,6 +20,113 @@ import (
...
+func (s *Sia) releaseDiskUsage(size int64) {
+	if s.diskUsageLimit == 0 {
+		return
+	}
+
+	s.diskUsageMu.Lock()
+	defer s.diskUsageMu.Unlock()
+	if size > 0 {
+		if uint64(size) > s.diskUsage {
+			s.diskUsage = 0
```
ChrisSchinnerl, line 75: We should never release more than we have previously acquired right? So I'd probably log a warning here if that happens to make sure we are aware of something going wrong.

ChrisSchinnerl (changes requested):

sia/objects.go
```diff
@@ -343,6 +453,25 @@ func (s *Sia) PutObject(ctx context.Context, accessKeyID string, bucket, object
...
+	defer func() {
+		if err == nil {
+			return
+		}
+		if objPath != "" {
+			if removeErr := os.Remove(objPath); removeErr != nil && !errors.Is(removeErr, os.ErrNotExist) {
+				s.logger.Error("failed to remove pending upload file after error",
+					zap.String("path", objPath),
+					zap.Error(removeErr))
+				return
```
peterjan: If this fails you skip `releaseDiskUsage`... is that intentional?

sia/objects_test.go
```diff
@@ -1036,6 +1037,173 @@ func TestOverwritePendingObjectCleansUpFile(t *testing.T) {
 	}
 }
 
+func TestDiskUsageLimit(t *testing.T) {
```
peterjan, line 1040: These are great but I would extend it with cases that cover `CopyObject` and `UploadPartCopy`, perhaps you can extend `TestDiskUsageLimitOngoingMultipartUpload` for the part copy one. It's important enough to warrant a test path for all three cases imo.

peterjan (changes requested):

sia/objects.go
chris124567: Yes it was a copilot review basically the idea was that you shouldn't decrease the amount of pending data on disk if you didn't actually remove the data

peterjan (dismissed):

sia/objects.go
```diff
@@ -20,6 +20,116 @@ import (
...
+		}
+	}
+	close(s.diskUsageWake)
+	s.diskUsageWake = make(chan struct{})
+}
+
+// removeAndRelease removes path and releases its bytes from the reserved
+// disk usage. The release only happens if the removal succeeds so that a
+// failed removal does not undercount disk usage.
+func (s *Sia) removeAndRelease(path string) error {
```
ChrisSchinnerl: This really adds a lot of surface for inconsistencies because it adds the files on disk as a separate source of truth in addition to our database. Instead I would get rid of `removeAndRelease` as well as `diskUsage` and do the following:

- In `UploadPart` return the size of the previous upload and call `releaseDiskUsage` with that number after attempting to remove the file from disk. Do so even if it fails to make sure uploads don't get stuck
- Same thing in `UploadPartCopy`
- In `lockUpload` replace `removeAndRelease` with `os.RemoveAll` and release the disk usage separately after `MarkObjectUploaded`.

Then the database is our only source of truth and not the data on disk. So if we fail to delete a file from disk or a new upload part gets snuck into our multipart upload folder while we finalize it, that's not an issue.

ChrisSchinnerl (changes requested):

peterjan (changes requested):
I was going to approve it and suggest the following, but then noticed that `notifyDiskUsage` is only called from `multipart.go`, aren't we missing some spots in objects.go that also need to notify? 

My suggestion was going to be the following little cleanup: if we create an `orphan` or `file` struct that holds filename and size, we can have our store methods return a pointer to that instead of 2,3 and in some cases 4 return variables by now. The call sites could then handle that file (new method on Sia) that combines `removeUpload`, `releaseDiskUsage` and maybe even `notifyDiskUsage` (which is how I noticed it's not being called in all places)

sia/persist/sqlite/objects.go
```diff
@@ -168,16 +167,16 @@ func getObject(tx *txn, obj *objects.Object, bid int64, name string, partNumber
 // remaining references, it is inserted into the orphaned_objects table. If
 // the overwrite leaves a previously pending file unreferenced, its filename
 // is returned so the caller can remove it from disk.
-func (s *Store) PutObject(accessKeyID, bucket, name string, contentMD5 [16]byte, meta map[string]string, length int64, fileName *string) (orphaned *string, size int64, _ error) {
+func (s *Store) PutObject(accessKeyID, bucket, name string, contentMD5 [16]byte, meta map[string]string, length int64, fileName *string) (orphan *objects.OrphanedFile, _ error) {
```
peterjan: `nit` update docstring

sia/persist/sqlite/objects.go
```diff
@@ -431,14 +430,14 @@ func (s *Store) ObjectsForUpload() ([]objects.ObjectForUpload, error) {
 // prior row is deleted first. If the prior row's sia_object_id is no longer
 // referenced, it is added to orphaned_objects. If the prior row's filename is
 // no longer referenced, it is returned for cleanup.
-func putObject(tx *txn, bid int64, name string, contentMD5 [16]byte, meta map[string]string, length int64, fileName *string, siaObject *objects.SiaObject) (*string, int64, error) {
+func putObject(tx *txn, bid int64, name string, contentMD5 [16]byte, meta map[string]string, length int64, fileName *string, siaObject *objects.SiaObject) (*objects.OrphanedFile, error) {
```
peterjan: `nit` update docstring

sia/multipart.go
```diff
@@ -395,19 +426,14 @@ func (s *Sia) CompleteMultipartUpload(ctx context.Context, accessKeyID, bucket,
...
-			s.logger.Warn("failed to remove pending upload file",
-				zap.String("bucket", bucket),
-				zap.String("object", object),
-				zap.String("filename", *orphaned),
-				zap.Error(err))
-		}
-	}
+	// wake any UploadPart waiting on the disk usage limit; HasMultipartUpload
+	// will now return ErrNoSuchUpload
+	s.releaseDiskUsage(0)
```
peterjan: why do you release here?

sia/multipart.go
```diff
@@ -102,27 +102,48 @@ func (s *Sia) AbortMultipartUpload(ctx context.Context, accessKeyID, bucket, obj
...
+	// allow exceeding the limit once any part has been stored so
+	// concurrent parts of the same upload can complete
+	allowExcess := func() (bool, error) {
+		return s.store.HasMultipartUpload(bucket, object, uploadID)
+	}
+	if err := s.addDiskUsage(ctx, opts.ContentLength, allowExcess); err != nil {
+		return nil, err
+	}
+	var partPath string
+	defer func() {
```
peterjan, line 131: can't you create an orphan here with partPath and opts.ContentLength and defer the cleanupOrphan? This was one of the places that triggered me to make the suggestion in the first place

sia/multipart.go
```diff
@@ -199,25 +209,29 @@ func (s *Sia) UploadPart(ctx context.Context, accessKeyID, bucket, object string
...
 
 	// add multipart part to the database
-	previous, err := s.store.AddMultipartPart(bucket, object, uploadID, filepath.Base(partPath), opts.PartNumber, contentMD5, contentLength)
+	previous, prevSize, err := s.store.AddMultipartPart(bucket, object, uploadID, filepath.Base(partPath), opts.PartNumber, contentMD5, contentLength)
 	if err != nil {
 		return nil, fmt.Errorf("failed to add part: %w", err)
-	} else if previous != "" {
+	}
+	// wake any concurrent UploadPart waiting on the disk usage limit
+	s.releaseDiskUsage(0)
```
peterjan: same here? calling `cleanupOrphan(orphan{filename: filepath.Join(partDir, previous)})` would do the trick right

peterjan (changes requested):

sia/multipart.go
chris124567: So concurrently uploading parts will re-evaluate `allowExcess`

peterjan (dismissed):

sia/multipart.go
```diff
@@ -395,19 +395,11 @@ func (s *Sia) CompleteMultipartUpload(ctx context.Context, accessKeyID, bucket,
...
-	if orphaned != nil {
-		if err := s.removeUpload(*orphaned); err != nil {
-			s.logger.Warn("failed to remove pending upload file",
-				zap.String("bucket", bucket),
-				zap.String("object", object),
-				zap.String("filename", *orphaned),
-				zap.Error(err))
-		}
-	}
+	s.cleanupOrphan(s.resolveOrphanPath(orphan))
```
ChrisSchinnerl: This makes for a pretty awkward pattern. Sometimes the store returns an `OrphanedFile`, other times it doesn't. Sometimes an `OrphanedFile` contains a relative and other times an absolute path if it has been resolved already. Sometimes we can just pass this to `cleanupOrphan` and other times we need to check for an empty string and then call `cleanupOrphan(nil)` etc.

It feels like adding the concept of an `OrphanedFile` saved us 1 return value but added a lot of magic that the compiler doesn't help us with. 

The logic might get more straightforward if we just get rid of that helper type and deal with one string and one number directly. Making it very clear that we are getting filenames back and then always resolving them to an appropriate absolute path within the context of whatever function we are calling `cleanupOrphan` from.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

peterjan (approved):

PR #141, opened 2026-04-28
Fix PartsCount

peterjan:
When retrieving an object by part number, the parts count was incorrectly returned as 0 for non-multipart objects, and the part query was missing filename, sia_object, and cached_at fields which prevented serving part data from disk or Sia.

### DIFF
--- a/sia/objects.go
+++ b/sia/objects.go
@@ -231,6 +231,10 @@ func (s *Sia) headOrGetObject(ctx context.Context, accessKeyID *string, bucket, 
 
 	var resp *s3.Object
 	if partNumber != nil {
+		partsCount := obj.PartsCount
+		if partsCount == 0 {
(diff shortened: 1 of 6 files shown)


chris124567 (dismissed):

sia/objects.go
```diff
@@ -231,14 +231,18 @@ func (s *Sia) headOrGetObject(ctx context.Context, accessKeyID *string, bucket,
 
 	var resp *s3.Object
 	if partNumber != nil {
+		partsCount := obj.PartsCount
+		if partsCount == 0 {
```
chris124567: nit `partsCount := min(obj.PartsCount, 1)`

chris124567 (approved):

ChrisSchinnerl (approved):

PR #144, opened 2026-05-07
Run compat tests against zeus

### LINKED ISSUE #50: Run the compatibility tests against the Sia backend

peterjan:
This PR extends our CI tests with a production Tox test running against Zeus. 

lt requires the following GH secrets:
- TS_OAUTH_CLIENT_ID
- TS_OAUTH_CLIENT_SECRET
- S3D_APP_KEY
- S3D_INDEXER_URL

I created an app key on zeus using the ultimate quota for now, perhaps we should add a custom quota though.

@ChrisSchinnerl mind adding the secrets?  The indexer URL should be https://siastorage.zeus.sia.dev the app key should be c20f77951e1725f583d7dcbe15f94f9e5320ff86a0ac4dbd185a116acbaf9df851f2835ac15307aa1825a95fb9c9c97611b1828dc3bc38d7d6e9b0c67427ead7 and the tailscale client ID and secret still need to be set up. I think the tailscale client will need a `tag:ci` tag with ACL access to the indexer.
                                                                                                         
Closes https://github.com/SiaFoundation/s3d/issues/50

### DIFF
--- a/.github/workflows/tox-test.yml
+++ b/.github/workflows/tox-test.yml
@@ -1,6 +1,8 @@
 name: S3 Compatibility Tests
 
 on:
+  push:
+    branches: [master]
   pull_request:
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


peterjan:
Not doing this after all

PR #146, opened 2026-05-08
Remove orphaned uploads

### LINKED ISSUE #121: Cleanup orphaned packed objects
We should have a background loop to see if there are files on disk that are not referenced by the database, remove those files and log a warning.

peterjan:
Closes https://github.com/SiaFoundation/s3d/issues/121


sia/sia.go
```diff
@@ -255,6 +258,72 @@ func (s *Sia) ProcessOrphans(ctx context.Context) {
...
+	defer t.Stop()
+
+	s.DeleteOrphanedUploads()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-t.C:
+			s.DeleteOrphanedUploads()
```
ChrisSchinnerl: Is there any need to run this in a loop? In theory we shouldn't produce orphans at runtime since all methods clean up after themselves. If we created orphans due to a crash then deleting them at startup should do the trick. 

Doing this at runtime also means you might introduce a race where a file exists on disk but not in the database yet. Resulting in the file getting deleted.

sia/persist/sqlite/objects.go
```diff
@@ -308,6 +308,44 @@ func (s *Store) SetObjectsCursor(cursor slabs.Cursor) error {
 	})
 }
 
+// AllFilenames returns the set of all filenames from the objects table and
+// in-progress multipart uploads.
+func (s *Store) AllFilenames() (map[string]struct{}, error) {
+	refs := make(map[string]struct{})
+	err := s.transaction(func(tx *txn) error {
+		objects, err := tx.Query(`SELECT DISTINCT filename FROM objects WHERE filename IS NOT NULL`)
```
ChrisSchinnerl: The filenames should already be unique right? Otherwise they couldn't exist in the same directory. So you should be able to return a slice here.

sia/upload_test.go
```diff
@@ -232,3 +233,89 @@ func TestOpenAndRemoveUpload(t *testing.T) {
...
+
+	// assert old orphaned multipart directory is removed
+	if _, err := os.Stat(oldOrphanDir); !os.IsNotExist(err) {
+		t.Fatal("old orphaned multipart directory should be removed")
+	}
+
+	// assert fresh orphaned file is kept
+	if _, err := os.Stat(freshOrphanPath); err != nil {
+		t.Fatal("fresh orphaned file should not be removed")
+	}
```
ChrisSchinnerl: might as well also check a fresh orphaned multipart upload dir.

ChrisSchinnerl (changes requested):

sia/sia.go
peterjan: I did this mostly to warn us of any file leaks in case we missed a spot, like the PR we have open. But fair point, in theory we shouldn't produce these at runtime. I'll remove the loop, I think the race was covered by the threshold and by reading from database first.

sia/persist/sqlite/objects.go
peterjan: Hm yeah sure, I don't like that it's not an explicit database constraint and we have to turn it into a map on the call site anyway.

chris124567 (changes requested):
fix conflicts

sia/sia.go
```diff
@@ -265,6 +266,44 @@ func (s *Sia) LoadSecret(ctx context.Context, accessKeyID string) (auth.SecretAc
...
+		return
+	}
+
+	// build lookup table
+	lookup := make(map[string]struct{})
+	for _, filename := range filenames {
+		lookup[filename] = struct{}{}
+	}
+
+	// remove unreferenced files older than the threshold
```
ChrisSchinnerl: There is no more threshold.

sia/persist/sqlite/objects.go
```diff
@@ -320,6 +320,31 @@ func (s *Store) SetObjectsCursor(cursor slabs.Cursor) error {
 	})
 }
 
+// AllFilenames returns all filenames from the objects table and in-progress
+// multipart uploads.
+func (s *Store) AllFilenames() (filenames []string, err error) {
+	err = s.transaction(func(tx *txn) error {
```
ChrisSchinnerl, line 326: `filenames` should be reset before the query here to avoid issues on retry

sia/sia_test.go
```diff
@@ -231,3 +232,98 @@ func newTestObject() sdk.Object {
...
+
+	// helper to assert file removal
+	assertRemoved := func(name string) {
+		t.Helper()
+		if _, err := os.Stat(filepath.Join(uploadsDir, name)); !os.IsNotExist(err) {
+			t.Fatalf("expected %q to be removed", name)
+		}
+	}
+
+	// add two objects with files on disk
```
ChrisSchinnerl: I don't think this is true? If you pass `true` to `createObject` you create the file on disk. 
So you create one object with a file on disk and one without. Then you do the same for multipart uploads.
`DeleteOrphanedUploads` then doesn't delete anything and finally you assert your test setup.

ChrisSchinnerl (changes requested):

sia/persist/sqlite/objects.go
peterjan: dammit, nice, we need a linter for this 😬

sia/sia_test.go
```diff
@@ -231,3 +232,103 @@ func newTestObject() sdk.Object {
...
+
+	// helper to assert file removal
+	assertRemoved := func(name string) {
+		t.Helper()
+		if _, err := os.Stat(filepath.Join(uploadsDir, name)); !os.IsNotExist(err) {
+			t.Fatalf("expected %q to be removed", name)
+		}
+	}
+
+	// add two objects, only one is on disk
```
ChrisSchinnerl: There might have been a misunderstanding. I pushed a commit to make it clear what I meant.

ChrisSchinnerl (changes requested):

sia/sia_test.go
peterjan: Thanks for that!

sia/sia.go
```diff
@@ -160,8 +161,13 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
...
 
-	// TODO: clean up orphaned uploads and multipart uploads in uploads
-	// directory on startup
+	// clean up any orphaned uploads on startup
+	deleted, err := sia.deleteOrphanedUploads()
+	if err != nil {
+		return nil, fmt.Errorf("failed to delete orphaned uploads: %w", err)
+	} else if deleted > 0 {
+		sia.logger.Info("removed orphaned uploads", zap.Int("removed", deleted))
+	}
```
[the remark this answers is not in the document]
chris124567: see https://github.com/SiaFoundation/s3d/pull/146#discussion_r3219320009

chris124567 (approved):
LGTM I agree with the Copilot nits though

ChrisSchinnerl (approved):

PR #149, opened 2026-05-11
Improve onboarding

### LINKED ISSUE #148: Improve Onboarding
Right now the access key id and secret keys need to be manually entered when running `s3d config`. It would be nicer if a user could leave it empty to generate one and have that then printed to stdout to copy and paste. 

We should also prompt for the recovery phrase during base interactive setup and do the same thing. If provided, use that, if not, generate one.
Creating a new one implicitly when running `s3d` for the first time seems like something that shouldn't happen after running `s3d config`. It's better to fail like we do for a missing seed in other software.

peterjan:
Closes https://github.com/SiaFoundation/s3d/issues/148

### DIFF
--- a/cmd/s3d/config.go
+++ b/cmd/s3d/config.go
@@ -118,14 +118,14 @@ func runConfigCmd(fp string) {
 
 	fmt.Println("")
 	if cfg.RecoveryPhrase != "" {
-		fmt.Println(ansiStyle("33", "A seed phrase is already configured."))
-		fmt.Println("If you change your seed phrase, you will have to re-register the app.")
-		if promptYesNo("Would you like to change your seed phrase?") {
+		fmt.Println(ansiStyle("33", "A recovery phrase is already configured."))
+		fmt.Println("If you change your recovery phrase, you will have to re-register the app.")
+		if promptYesNo("Would you like to change your recovery phrase?") {
 			cfg.RecoveryPhrase = ""
-			setSeedPhrase()
+			setRecoveryPhrase()
 		}
 	} else {
-		setSeedPhrase()
+		setRecoveryPhrase()
 	}
 
 	fmt.Println("")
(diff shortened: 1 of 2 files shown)


cmd/s3d/config.go
```diff
@@ -268,6 +309,48 @@ func setListenAddress(context string, value *string) {
...
+			fmt.Println("")
+
+			for {
+				confirm := readPasswordInput("Confirm seed phrase")
+				if confirm == phrase {
+					break
+				}
+				fmt.Println(ansiStyle("31", "Seed phrases do not match!"))
+				fmt.Println(ansiStyle("31", fmt.Sprintf("Expected: %q", phrase)))
+				fmt.Println(ansiStyle("31", fmt.Sprintf("Entered:  %q", confirm)))
```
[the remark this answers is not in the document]
peterjan: Printing both because it's useful to see the mismatch in case you fat-fingered a word. We already print it anyway and hostd and indexd do the same thing.

cmd/s3d/config.go
```diff
@@ -112,6 +116,18 @@ func runConfigCmd(fp string) {
 	fmt.Println("")
 	setDataDirectory()
 
+	fmt.Println("")
+	if cfg.RecoveryPhrase != "" {
+		fmt.Println(ansiStyle("33", "A seed phrase is already configured."))
```
ChrisSchinnerl: Since we call this recovery phrase in Sia Storage we should probably be consistent and call it that everywhere.

cmd/s3d/config.go
```diff
@@ -268,6 +309,48 @@ func setListenAddress(context string, value *string) {
 	}
 }
 
+func setSeedPhrase() {
```
ChrisSchinnerl: Recovery phrase

ChrisSchinnerl (changes requested):

cmd/s3d/config.go
peterjan: Typing "seed" to generate didn't make sense any more so I use "leave blank to.." which is consistent with how we generate access keys.

ChrisSchinnerl (approved):

PR #150, opened 2026-05-11
Handle `If-Modified-Since`

peterjan:
Found out `If-Modified-Since` was being ignored considering we compare it against `obj.Metadata["Last-Modified"]` which never gets populated.

### DIFF
--- a/s3/objects.go
+++ b/s3/objects.go
@@ -953,6 +953,7 @@ func writeGetOrHeadObjectHeaders(obj *Object, w http.ResponseWriter, r *http.Req
 
 	etag := FormatETag(obj.ContentMD5[:], 0)
 	w.Header().Set("ETag", etag)
+	w.Header().Set("Last-Modified", obj.LastModified.UTC().Format(http.TimeFormat))
 
 	if r.Header.Get("If-None-Match") == etag {
 		return s3errs.ErrNotModified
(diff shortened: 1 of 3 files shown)


s3/objects.go
```diff
@@ -962,9 +963,8 @@ func writeGetOrHeadObjectHeaders(obj *Object, w http.ResponseWriter, r *http.Req
 		w.Header().Set("x-amz-mp-parts-count", fmt.Sprintf("%d", *obj.PartsCount))
 	}
 
-	lastModified, _ := time.Parse(http.TimeFormat, obj.Metadata["Last-Modified"])
 	ifModifiedSince, _ := time.Parse(http.TimeFormat, r.Header.Get("If-Modified-Since"))
```
chris124567: A recipient [MUST](https://datatracker.ietf.org/doc/html/rfc7232) ignore If-Modified-Since if the request contains an If-None-Match header field; the condition in If-None-Match is considered to be a more accurate replacement for the condition in If-Modified-Since, and the two are only combined for the sake of interoperating with older intermediaries that might not implement If-None-Match.

```
if r.Header.Get("If-None-Match") == "" {
  ifModifiedSince, _ := time.Parse(http.TimeFormat, r.Header.Get("If-Modified-Since"))
  if !ifModifiedSince.IsZero() && !ifModifiedSince.Before(obj.LastModified) {
    return s3errs.ErrNotModified
  }
}
```

ChrisSchinnerl (approved):

PR #153, opened 2026-05-13
Add `s3d login` command

### LINKED ISSUE #152: Add `s3d login` command
Due to the nature of the Sia Storage login it's a bit awkward to store a recovery phrase in the config since we only need it once and then store the app key in the database.

So we should remove it from the config file and introduce an `s3d login` command a user can run once to log in and then run `s3d`.

If you don't have a config yet `s3d login` would first act like `s3d config` followed by the login. If you already have one it just logs you in.

chris124567:
Close #152


cmd/s3d/login.go
```diff
@@ -0,0 +1,122 @@
...
+			fmt.Println("")
+
+			for {
+				confirm := readPasswordInput("Confirm recovery phrase")
+				if confirm == phrase {
+					return phrase
+				}
+				fmt.Println(ansiStyle("31", "Recovery phrases do not match!"))
+				fmt.Println(ansiStyle("31", fmt.Sprintf("Expected: %q", phrase)))
+				fmt.Println(ansiStyle("31", fmt.Sprintf("Entered:  %q", confirm)))
```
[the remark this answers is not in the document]
chris124567: These were preexisting

README.md
```diff
@@ -46,16 +46,24 @@ which will guide you through an interactive configuration process and generate a
 config file. For more information on configuration options, see the
 [Configuration](#configuration) section below.
 
-Once configured, start `s3d` with
+Once configured, log in to register `s3d` with the Sia indexer with
 
 ```sh
-./s3d
+s3d login
 ```
```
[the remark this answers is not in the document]
chris124567: fixed

README.md
```diff
@@ -46,16 +46,24 @@ which will guide you through an interactive configuration process and generate a
...
+URL in your browser to complete setup.
+
+If no config file exists yet, `s3d login` will run the configuration wizard
+first.
+
+Once logged in, start `s3d` with
+
+```sh
+./s3d
+```
```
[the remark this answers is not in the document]
chris124567: fixed

cmd/s3d/config.go
```diff
@@ -64,12 +63,11 @@ type (
...
-		RecoveryPhrase string `yaml:"recoveryPhrase"`
-		Directory      string `yaml:"directory"`
-		Log            Log    `yaml:"log"`
-		S3             S3     `yaml:"s3"`
-		Sia            Sia    `yaml:"sia"`
+		ApiAddress string `yaml:"apiAddress"`
+		Directory  string `yaml:"directory"`
+		Log        Log    `yaml:"log"`
+		S3         S3     `yaml:"s3"`
+		Sia        Sia    `yaml:"sia"`
```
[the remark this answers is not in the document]
chris124567: Anyone feel strongly about leaving in a deprecated field or setting KnownFields to false?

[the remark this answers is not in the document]
peterjan: Still early enough to break this imo

cmd/s3d/login.go
```diff
@@ -0,0 +1,122 @@
...
+		ServiceURL:  "https://github.com/SiaFoundation/s3d",
+	})
+}
+
+func runLoginCmd(ctx context.Context, configPath string) {
+	// if no config exists yet, run the config wizard first
+	if configPath == "" {
+		fmt.Println("No existing config found. Launching configuration wizard.")
+		fmt.Println("")
+		runConfigCmd("")
```
peterjan, line 39: `nit` I would instruct the user to run the command, not run it here to then continue with the login command after config is finished. I haven't tried it but I think that would be kind of confusing?

cmd/s3d/login.go
```diff
@@ -0,0 +1,122 @@
...
+	fmt.Println("This command will register s3d with the Sia indexer.")
+	fmt.Println("You will be prompted for a recovery phrase and asked to visit a URL to approve the app connection.")
+	fmt.Println("")
+
+	store, err := openStore(zap.NewNop())
+	checkFatalError("failed to open database", err)
+	defer store.Close()
+
+	if _, err := store.AppKey(); err == nil {
+		fmt.Println(ansiStyle("33", "s3d is already registered for this data directory."))
```
peterjan: It might be useful to print the app key? Not too sure about that though but it might be useful to know the app key so you know exactly which one you are un-registering

cmd/s3d/login.go
```diff
@@ -0,0 +1,122 @@
...
+		fmt.Println("If you log in again, you will re-register the app.")
+		if !promptYesNo("Would you like to log in again?") {
```
peterjan: `nit` I would drop the line above and change this to "Would you like to re-register the app" to avoid using "login" and "register" terminology that essentially mean the same thing. Later on we also talk about "app connections", perhaps it's too late and it would be inconsistent with other SDKs and login procedures but it would probably be better to drop the word "connection" in all those phrases.

cmd/s3d/login.go
```diff
@@ -0,0 +1,122 @@
...
+	checkFatalError("failed to wait for app approval", err)
+	if !approved {
+		fmt.Println(ansiStyle("31", "app connection was declined"))
+		os.Exit(1)
+	}
+
+	sdkClient, err := builder.Register(ctx, phrase)
+	checkFatalError("failed to register app", err)
+
+	checkFatalError("failed to store app key", store.SetAppKey(sdkClient.AppKey()))
```
peterjan: I wonder if we should clear the app key as soon as the user says he wants to re-register the app... I think that's kind of nice because if you timeout the app connection or cancel/fail it, you don't get asked the same question if you run the command again but instead get taken to the login flow immediately.

cmd/s3d/login.go
ChrisSchinnerl: I don't think that would do us any good. There is nothing to compare it to anyway. We could probably print the indexer you are logged in to though. Since app keys are unique to a recovery phrase + indexer.

cmd/s3d/login.go
ChrisSchinnerl: I would print `This app is already registered with <indexer url>.` and then exit. 
We can follow up without logout functionality.

cmd/s3d/login.go
ChrisSchinnerl: Let's ignore logout for now. There are a ton of implications to it and it probably deserves a separate PR.

cmd/s3d/config.go
ChrisSchinnerl: Let's remove this from the config as well. Your app key is a result of an indexer + recovery phrase. So when you log in we can persist that alongside the app key and when we a user tries to log in again we can tell them what indexer they are logged in with.

cmd/s3d/login.go
ChrisSchinnerl: This is intentional. The idea being that `s3d login` is really all you need to get started. You only need `s3d config` then to overwrite the config later on.

ChrisSchinnerl (changes requested):

cmd/s3d/login.go
peterjan: I think Sia storage should probably show the app key in the interface because if I register two apps with the same metadata I have no clue which is which. (Not that you would do that :p but still)

cmd/s3d/login.go
peterjan: I still think it's weird but ok. I understand that it would not be consistent with our other daemons, but if that is the idea I think it should be a single `init` command that takes you through the entire setup, configuring and registering.

cmd/s3d/login.go
```diff
@@ -0,0 +1,119 @@
...
+	checkFatalError("failed to open database", err)
+	defer store.Close()
+
+	if _, existingURL, err := store.AppKey(); err == nil {
+		fmt.Println(ansiStyle("33", fmt.Sprintf("This app is already registered with %s.", existingURL)))
+		return
+	} else if !errors.Is(err, sqlite.ErrNoAppKey) {
+		checkFatalError("failed to check app key", err)
+	}
+
```
ChrisSchinnerl, line 57: Prompt the indexer first with `sia.storage` as the default.

README.md
ChrisSchinnerl: We should update this now to tell users to use `s3d login` which will guide them through the configuration and link their app to their indexer. Also that they can update the configuration later by calling `s3d config`.

cmd/s3d/main.go
```diff
@@ -163,49 +172,17 @@ func main() {
...
-		if err != nil {
-			log.Fatal("failed to register app", zap.Error(err))
-		}
-		if err := store.SetAppKey(sdkClient.AppKey()); err != nil {
-			log.Fatal("failed to store app key in database", zap.Error(err))
-		}
-	} else {
+	appKey, indexerURL, err := store.AppKey()
+	if errors.Is(err, sqlite.ErrNoAppKey) {
+		checkFatalError("No app key found. Please run 's3d login' to register the app", err)
```
ChrisSchinnerl: If you print the error here it results in `No app key found. Please run 's3d login' to register the app: no app key set` which is a bit much.
```suggestion
		checkFatalError("No app key found. Please run 's3d login' to register the app")
```

ChrisSchinnerl (changes requested):

docker-compose.yml
```diff
@@ -2,11 +2,10 @@
 #
 # Before starting the service, run the configuration wizard:
 #   docker compose run --rm -it s3d config
+# Then register s3d with the Sia indexer:
```
peterjan: Should we update

```
# Before starting the service, run the configuration wizard:
#   docker compose run --rm -it s3d config
```

now that s3d login covers both?

README.md
```diff
@@ -89,19 +94,18 @@ Run the configuration wizard to generate a config file.
 docker compose run --rm s3d config
 ```
 
-### 3. Start `s3d`
+### 3. Log in
+
+Register `s3d` with the indexer.  `s3d` will print a URL that must be visited to approve the connection to the indexer.
```
peterjan:
```suggestion
Register `s3d` with the indexer. `s3d` will print a URL that must be visited to approve the connection to the indexer.
```

peterjan (dismissed):
`s3d init` covers the load better imo, but leaving that up to

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

PR #154, opened 2026-05-14
Update README

peterjan:
Leftover from PRs we merged recently, forgot to update the README


peterjan:
I just found out about https://github.com/SiaFoundation/s3d/issues/152 which changes this again anyway

PR #168, opened 2026-05-26
Add user support

### LINKED ISSUE #156: Add support for users
Something we already have in the memory backend but not the Sia backend is users. A user owns buckets they create and access keys.
A user can access only their own files in their own buckets by authenticating with their own keys.

For this issue to be closed we need:

- update the database and codebase to support users and have them own keys
- enable any potentially disabled compat tests
- `s3d users` commands for creating/deleting/listing users and generating/deleting/listing keypairs for them by interacting with the database
- remove the `keypairs` from the config

peterjan:
Closes https://github.com/SiaFoundation/s3d/issues/156


sia/persist/sqlite/migrations_test.go
```diff
@@ -16,12 +16,27 @@ const initialSchema = `/*
...
+    FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE
+);
+CREATE INDEX access_keys_user_id_idx ON access_keys(user_id);
 
 CREATE TABLE buckets (
     id INTEGER PRIMARY KEY AUTOINCREMENT,
     created_at INTEGER NOT NULL,
-    name TEXT NOT NULL UNIQUE
+    name TEXT NOT NULL UNIQUE,
+    user_id INTEGER NOT NULL,
```
peterjan, line 36: I've been adding migrations since we were public, but this one is a little annoying. We could add a default user, which isn't the worst thing... On the other hand, we're also changing the config, so if we write a migration we should probably also migrate the config, which I wanted to avoid. Perhaps we can get away without migration considering we still have the big warning in the README?

sia/sia_test.go
```diff
@@ -205,13 +205,27 @@ func NewTester(t testing.TB, opts ...testutil.TesterOption) *testutil.S3Tester {
 	}
 	t.Cleanup(func() { store.Close() })
 
+	// create test user and access key
+	if err := store.CreateUser(testutil.Owner); err != nil {
```
chris124567: nit: These calls are redundant given that we call `NewCustomTester`

chris124567 (changes requested):
Fix conflicts

sia/sia_test.go
peterjan: woopsie poepsie

s3/buckets.go
chris124567, line 1: Not super sensitive but we should check that we are the bucket owner in `bucketLocation` by calling HeadBucket or something similar

Otherwise LGTM I tested with s3 CLI and everything seemed to work correctly

peterjan:
will be 🟢  after https://github.com/SiaFoundation/s3-tests/pull/9

cmd/s3d/keys.go
```diff
@@ -0,0 +1,134 @@
...
+		fmt.Fprintln(os.Stderr, keysUsage)
+		os.Exit(1)
+	}
+
+	switch args[0] {
+	case "create":
+		fs := flag.NewFlagSet("keys create", flag.ExitOnError)
+		accessKeyFlag := fs.String("access-key", "", "access key ID (auto-generated if empty)")
+		secretKeyFlag := fs.String("secret-key", "", "secret key (auto-generated if empty)")
+		fs.Parse(args[1:])
```
ChrisSchinnerl: Is there a reason you are doing this in such a "manual" way instead of making use of the `Sub` commands when calling `flagg.Parse`? This splits up the command line parsing into multiple files when it could all be in one spot. The actual implementations for each command could still be in separate files but right now this reimplements parsing for sub commands.

cmd/s3d/main.go
```diff
@@ -165,9 +175,15 @@ func main() {
...
 
-	// before initializing the SDK, check whether we have at least one key pair configured
-	if len(cfg.Sia.KeyPairs) == 0 {
-		checkFatalError("Please provide at least one valid key pair. You can do so by updating the config file or running the 'config' command", sia.ErrNoAccessKey)
+	// check that at least one access key is configured
+	keys, err := store.ListAccessKeys(nil)
+	if err != nil {
+		checkFatalError("failed to list access keys", err)
+	} else if len(keys) == 0 {
+		os.Stderr.WriteString("No access keys configured. Create a user and key pair first:\n\n")
```
ChrisSchinnerl: I don't think this should be a reason for failure. It takes us back to requiring users to running more than 1 command before s3d starts. 

I'd rather prompt them to optionally create a user and key during config, tell them how they can do it later and allow them to start without any keys.

sia/persist/sqlite/init.sql
```diff
@@ -3,12 +3,27 @@
...
+);
+CREATE INDEX access_keys_user_id_idx ON access_keys(user_id);
 
 CREATE TABLE buckets (
     id INTEGER PRIMARY KEY AUTOINCREMENT,
     created_at INTEGER NOT NULL,
-    name TEXT NOT NULL UNIQUE
+    name TEXT NOT NULL UNIQUE,
+    user_id INTEGER NOT NULL,
+    FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE RESTRICT
```
ChrisSchinnerl:
```suggestion
    FOREIGN KEY (user_id) REFERENCES users(id)
```

sia/persist/sqlite/init.sql
```diff
@@ -3,12 +3,27 @@
 	migrations.go
 */
 
+CREATE TABLE users (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
```
ChrisSchinnerl: I don't think we need `AUTOINCREMENT` here? We probably also don't need it for the `buckets` honestly. Not sure why that got in there.

sia/persist/sqlite/migrations_test.go
ChrisSchinnerl: Tbh I don't know why you started doing them in the first place considering we haven't officially released this yet 😅

sia/persist/sqlite/objects.go
```diff
@@ -145,7 +145,7 @@ func getObject(tx *txn, obj *objects.Object, bid int64, name string, partNumber
 // remaining references, it is inserted into the orphaned_objects table. If
 // the overwrite leaves a previously pending file unreferenced, its filename
 // is returned so the caller can remove it from disk.
-func (s *Store) PutObject(accessKeyID, bucket, name string, contentMD5 [16]byte, meta map[string]string, length int64, fileName *string) (*string, error) {
+func (s *Store) PutObject(bucket, name string, contentMD5 [16]byte, meta map[string]string, length int64, fileName *string) (*string, error) {
```
ChrisSchinnerl: We still need to pass in all those access key ids to check bucket ownership under the same transaction that modifies or fetches an object. 

Maybe one way to achieve this would be to pass `accessKeyID` to `bucketID` and then `bucketID` only gives you the id of the bucket if you have access to it.

sia/persist/sqlite/objects.go
```diff
@@ -174,7 +174,7 @@ func (s *Store) MarkObjectUploaded(bucket, name string, contentMD5 [16]byte, sea
 		err = tx.QueryRow(`
 			UPDATE objects
 			SET sia_object_id = $1, sia_object = $2, filename = NULL
-			WHERE bucket_id = $4 AND name = $5 AND sia_object_id IS NULL
+			WHERE bucket_id = $3 AND name = $4 AND sia_object_id IS NULL
```
ChrisSchinnerl, line 198: This hasn't caused any issues so far?

ChrisSchinnerl (changes requested):

sia/persist/sqlite/migrations_test.go
peterjan: Making it public was probably a mistake. I just wanted to be nice to people that are already running it. From Discord etc I noticed that quite a few people are effectively using it already.

sia/persist/sqlite/init.sql
peterjan: It's just something that's in my fingers from MySQL/postgres. I never really stopped to think that it's not necessary in SQLite. I'll get rid of it.

sia/persist/sqlite/objects.go
peterjan: No it's pretty crazy, in sqlite `$N` is not a positional parameter but a named parameter, the driver allows it as long as there's 4 arguments being passed

sia/persist/sqlite/objects.go
peterjan: I wanted to avoid redundant auth checks in the store but that does indeed introduce a race, all be it small. Passing it to `bucketID` is a good idea though, I'll make that change

cmd/s3d/main.go
peterjan: Ok, makes sense

cmd/s3d/keys.go
peterjan: I was defining the commands in main and I didn't like the bulky switch case and all the flags/options all there in the main function. I actually thought it was cleaner to move that information into a separate file but then when you commented it splits up the command parsing I can obviously also see value in keeping that together. I don't know I just ended up with this, I added the access-key and secret-key flag afterwards, perhaps that's why it didn't trigger me until the very end...

cmd/s3d/keys.go
```diff
@@ -0,0 +1,132 @@
...
+
+	keys, err := store.ListAccessKeys(userName)
+	checkFatalError("failed to list access keys", err)
+
+	if len(keys) == 0 {
+		fmt.Println("No access keys found.")
+		return
+	}
+	for _, k := range keys {
+		fmt.Printf("%s\t%s\n", k.AccessKeyID, k.UserName)
```
ChrisSchinnerl: Might as well print the secret key considering that it is stored in the database and the user might as well just query that themselves at this point. But then we also need to update the "The secret key will not be shown again" string.

ChrisSchinnerl (changes requested):
Tested it and it works but the cli part feels a bit clunky. We can follow up with that but just to give you an example of what I mean.

The following gives no information about a potential command for user or key creation.
```
[cschinnerl:~/dev/s3d]$ s3d --help
Usage of s3d:
  -api.s3 string
        address to serve S3 API on (default "127.0.0.1:8000")
```

The following mistake causes s3d to start rather than return an error for unknown command
```
[cschinnerl:~/dev/s3d]$ s3d create
failed to start S3 API listener: listen tcp 127.0.0.1:8000: bind: address already in use
```

ChrisSchinnerl (approved):

PR #173, opened 2026-05-27
Support arbitrary delimiters in ListObjects

### LINKED ISSUE #169: Support arbitrary delimiters in ListObjects
ListObjectsV1|2 currently return 501 Not Implemented when the delimiter is anything other than /. While / is by far the most common delimiter, the S3 spec allows any single character. Some tools and SDKs use alternative delimiters (e.g., \ or custom separators for non-hierarchical key schemes). Returning a hard 501 breaks interoperability and is difficult for callers to diagnose since most S3 clients don't expect this error from a listing call. I think we already support arbitrary delimiters, not 100% sure why we restrict it to "/" only.

chris124567:
Close #169

```
goos: linux
goarch: amd64
pkg: github.com/SiaFoundation/s3d/sia/persist/sqlite
cpu: Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz
BenchmarkListObjects/no_delimiter_no_prefix-4                286           3916291 ns/op
BenchmarkListObjects/slash/root_delimiter-4                   20          52887625 ns/op
BenchmarkListObjects/slash/random_without_delimiter-4        816           1515163 ns/op
BenchmarkListObjects/slash/random_with_delimiter-4          1434            769890 ns/op
BenchmarkListObjects/slash/folder_bottom_delimiter-4        8109            166300 ns/op
BenchmarkListObjects/slash/folder_delimiter-4               1570            766823 ns/op
BenchmarkListObjects/backslash/root_delimiter-4               19          54393207 ns/op
BenchmarkListObjects/backslash/random_without_delimiter-4                    850           1631373 ns/op
BenchmarkListObjects/backslash/random_with_delimiter-4                      1550            786753 ns/op
BenchmarkListObjects/backslash/folder_bottom_delimiter-4                    7903            164151 ns/op
BenchmarkListObjects/backslash/folder_delimiter-4                           1610            740688 ns/op
PASS
ok      github.com/SiaFoundation/s3d/sia/persist/sqlite 80.354s
```

### DIFF
--- a/s3/objects.go
+++ b/s3/objects.go
@@ -456,10 +456,6 @@ func (s *s3) listObjectsV1(w http.ResponseWriter, r *http.Request, accessKeyID *
 		page.Marker = &marker
 	}
 
-	if prefix.Delimiter != "" && prefix.Delimiter != "/" {
-		return s3errs.ErrNotImplemented // only "/" delimiter is supported
-	}
-
 	// list objects
 	objects, err := s.backend.ListObjects(r.Context(), accessKeyID, bucket, prefix, page)
 	if err != nil {
@@ -505,8 +501,6 @@ func (s *s3) listObjectsV2(w http.ResponseWriter, r *http.Request, accessKeyID *
 	page, err := listObjectsPageFromQuery(q)
 	if err != nil {
 		return err
-	} else if prefix.Delimiter != "" && prefix.Delimiter != "/" {
-		return s3errs.ErrNotImplemented // only "/" delimiter is supported
 	}
 
 	// don't allow unordered listing with delimiter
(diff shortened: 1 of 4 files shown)


s3/objects.go
```diff
@@ -456,10 +456,6 @@ func (s *s3) listObjectsV1(w http.ResponseWriter, r *http.Request, accessKeyID *
 		page.Marker = &marker
 	}
 
-	if prefix.Delimiter != "" && prefix.Delimiter != "/" {
-		return s3errs.ErrNotImplemented // only "/" delimiter is supported
```
ChrisSchinnerl, line 460: Is there a compatibility test in the test suite we could enable now to cover this?

ChrisSchinnerl (approved):

peterjan (approved):
Can any of you remember why we disabled this? We already supported it but deliberately disabled support, at least that's how it looks to me. I remember writing the directory walk test and that worked with crazy separators... I feel there must've been a reason why decided to disable it?

PR #174, opened 2026-05-27
Make background upload pipeline state visible through endpoint

### LINKED ISSUE #171: Background upload pipeline state is invisible to operators
There is no way to query how many objects are pending upload, how many have failed, or what the current upload queue depth is. This information exists in the SQLite database but is not exposed. We could add a stats endpoint or a status command.

chris124567:
Close #171

Do we want to require auth for this?


s3/s3.go
```diff
@@ -228,6 +228,9 @@ type Backend interface {
 	//
 	// - If the last part is below the minimum size, [ErrEntityTooSmall] must be returned.
 	CompleteMultipartUpload(ctx context.Context, accessKeyID, bucket, object string, uploadID UploadID, parts []CompleteMultipartPart) (*CompleteMultipartUploadResult, error)
+
+	// UploadStats returns statistics about the background upload pipeline.
+	UploadStats(ctx context.Context) (UploadStats, error)
```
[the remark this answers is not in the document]
chris124567: fixed forgot to add file

s3/s3.go
```diff
@@ -414,7 +417,9 @@ func (s *s3) routeBase(w http.ResponseWriter, r *http.Request, accessKeyID *stri
 	// common headers at
 	// https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html.
 	//
-	if uploadID := query.Get("uploadId"); uploadID != "" {
+	if r.Method == http.MethodGet && path == "_s3d/status/uploads" {
+		err = s.handleUploadStats(w, r)
+	} else if uploadID := query.Get("uploadId"); uploadID != "" {
```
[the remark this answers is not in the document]
chris124567: fixed forgot to add file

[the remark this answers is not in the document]
chris124567: fixed

sia/persist/sqlite/objects_test.go
```diff
@@ -1352,6 +1352,88 @@ func TestObjectsForUpload(t *testing.T) {
 	}
 }
 
+func TestUploadStats(t *testing.T) {
```
peterjan, line 1431: The test doesn't cover the `FailedUploads` counter

peterjan (changes requested):
I opened the issue but I'm not sure how Chris or Nate feel about an endpoint such as this. I think if we go ahead and add it, it should definitely require authentication though. I'm wondering whether it makes more sense to expose it as a prometheus endpoint so an operator can easily graph the upload pipeline. I played around with s3d on my machine and I only have 40mbps up, so I really need the disk storage limit flag and if I were to run s3d on my machine I definitely need more insights into the upload progress... 

edit: requested changes only because we might want to consider prometheus endpoint and exposing it through a /metrics, PR looks good to me otherwise

sia/persist/sqlite/objects_test.go
chris124567: We don't have that information in the store

s3/s3.go
```diff
@@ -286,8 +299,21 @@ func New(b Backend, opts ...Option) http.Handler {
 		handler = s3.hostBucketBaseMiddleware(handler)
 	}
 
-	// authentication middleware
-	return s3.authMiddleware(handler)
+	return s3.statusMiddleware(s3.authMiddleware(handler))
```
ChrisSchinnerl: We should probably spin up that API on a separate port altogether. Similar to how indexd has an app and an admin api this would have an admin api and the s3 one. That gives you the extra security of being able to expose one of them to the internet and keep the other one local. 

This also allows us to more easily extend the admin api with a UI we serve or more routes without running into collisions.

Which means we should probably return 2 handlers here and then spin up 2 web servers.

ChrisSchinnerl (changes requested):

peterjan (changes requested):
I think we should call it an admin API and have /metrics be the first and for the time being only endpoint that serves the upload stats as prometheus metrics? I'd confirm this with either Chris or Nate but after our discussion I feel that was what we decided.

ChrisSchinnerl:
> I think we should call it an admin API and have /metrics be the first and for the time being only endpoint that serves the upload stats as prometheus metrics? I'd confirm this with either Chris or Nate but after our discussion I feel that was what we decided.

Yeah that sounds good. Maybe even make it `/prometheus` since we learned that doing both regular metrics + prometheus flag doesn't always work out.

peterjan (changes requested):
I think we should extend the README and have a section about the admin API. We should also be able to configure it through the config command. Right now the feature is really very hidden and disabled by default.

peterjan (dismissed):

cmd/s3d/config.go
```diff
@@ -44,6 +44,13 @@ type (
 	// S3 contains S3 related configuration.
 	S3 struct {
 		HostBases []string `yaml:"hostBases"`
+		// AdminAddress is the address the admin API is served on. It must
+		// differ from the S3 API address. If empty, the admin API is not
+		// started.
+		AdminAddress string `yaml:"adminAddress"`
+		// AdminPassword is the password required to access the admin API via
+		// HTTP Basic authentication. If empty, the admin API is inaccessible.
+		AdminPassword string `yaml:"adminPassword"`
```
ChrisSchinnerl: Let's move these under `Config` to the `ApiAddress`. It's a bit odd to have the admin api which in theory doesn't have anything to do with S3 in particular under `S3` but the actual S3 api in `Config`.

cmd/s3d/config.go
```diff
@@ -221,10 +228,54 @@ func setAdvancedConfig() {
...
 	fmt.Println("It should only be exposed to the public internet via an https reverse proxy")
 	setListenAddress("HTTP Address", &cfg.ApiAddress)
+
+	setAdminAPI()
+}
+
+func setAdminAPI() {
+	fmt.Println("")
+	fmt.Println("The admin API serves Prometheus metrics about the upload pipeline on a separate HTTP address.")
+	fmt.Println("It is disabled by default.")
```
ChrisSchinnerl: Let's always enable it. I imagine we might want to access it from the CI as well for users who don't want to set up a full prometheus+grafana stack. e.g. via a `s3d status` command.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

peterjan (approved):

PR #175, opened 2026-05-29
Implement streaming auth

### LINKED ISSUE #17: Implement streaming authentication
Check out `handleAuthV4`. We should be able to implement a chunk reader that handles all the cases starting with `ContentStreaming` similar to how GofakeS3 does it.

peterjan:
This PR adds implementations for `STREAMING-AWS4-HMAC-SHA256-PAYLOAD` and `STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER`, however most clients no longer emit these. Alternatively we could keep the 501 in place and return a more meaningful error perhaps. AWS CLI v1 and hand rolled clients might still use it, and by implementing it we're more feature complete... 

Closes https://github.com/SiaFoundation/s3d/issues/17


s3/auth/auth_v4_test.go
```diff
@@ -19,6 +31,16 @@ func (s *mockKeyStore) LoadSecret(_ context.Context, _ string) (SecretAccessKey,
 	return nil, s3errs.ErrInvalidAccessKeyId
 }
 
+type staticKeyStore map[string]SecretAccessKey
```
ChrisSchinnerl: I'd update the `mockKeyStore` instead of introducing an additional static one.

ChrisSchinnerl (changes requested):
Looking pretty good. Only left one small comment and added a test+fix to make sure we get some e2e testing with the minio client in addition to our own unit tests.

ChrisSchinnerl (approved):

PR #178, opened 2026-06-01
Minor fixes

ChrisSchinnerl:
This PR updates the format string for `ContentTime` to use `000Z` instead of `999Z` to make sure strict parsers don't fail. With `999Z` Go will omit the dot and 3 millisecond digits on whole-second timestamps. With `000Z` it always returns the full timestamp. 

The other one is the `slabs.ErrObjectNotFound` error that we can't check for with `errors.Is` since it is returned by an http API.


chris124567 (approved):

peterjan (approved):
the strings.Contains is quite sad

PR #184, opened 2026-06-05
Remove memory backend

### LINKED ISSUE #132: Remove memory backend
The memory backend was useful while the indexer and the sdk were under heavy development but now we just have a lot of almost duplicated logic.

Instead, we can make what we currently call the Sia backend the only backend but pass in an ephemeral/in-memory/mocked SDK to still get lightweight testing that doesn't require a full indexer.

chris124567:
Close #132


peterjan (dismissed):
Raked up some conflicts, otherwise LGTM

ChrisSchinnerl (dismissed):

ChrisSchinnerl (approved):

PR #192, opened 2026-06-08
Check remaining storage before upload

### LINKED ISSUE #188: Check account limits before uploading to Sia
In `uploadObjects` before we actually upload we should check the remaining amount of data a user can pin to their account and if it is insufficient, interrupt the upload and log a warning instead of uploading the data and failing to pin over and over again.

peterjan:
Closes https://github.com/SiaFoundation/s3d/issues/188


ChrisSchinnerl (approved):

sia/upload.go
```diff
@@ -220,8 +229,17 @@ func (s *Sia) uploadObjects(ctx context.Context) { //nolint:revive
 		})
 	}
 
-	// send uploads to workers
+	// send uploads to workers, skip groups that exceed the remaining
+	// storage space to avoid failed pin attempts after upload
 	for _, g := range groups {
+		if uint64(g.totalSize) > remaining {
```
ChrisSchinnerl, line 237: Can you count the groups and then print a single warning? Otherwise this is going to spam potentially a lot of lines every minute.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

PR #193, opened 2026-06-08
Decouple uploading and pinning

### LINKED ISSUE #189: Separate uploading from pinning
S3d will currently just retry failed uploads when pinning the object fails. Which is unfortunate because the metadata is valid, it's just the pinning that failed. 

To fix this we should decouple pinning from uploading. When an object is uploaded we should store the metadata in the database but not delete the local file yet. The pinning loop then fetches objects that need to be pinned and upon success deletes the local file.

NOTE:
- We hosts don't store the data forever without the object being pinned. So let's set a 24 hour deadline on pinning. After that it's back to reuploading the object. 
- Don't pin repeatedly if pinning an object fails. It should only retry once per hour or so if pinning fails the first time

ChrisSchinnerl:
This PR decouples pinning into a separate routine instead of synchronously doing it after uploading one or more objects. This has a few reasons.

1. Pinning is slow when uploading lots of tiny, packed files. To a point where uploads can be bottlenecked by pinning.
2. A temporary issue when pinning a file shouldn't cause you to waste all that bandwidth again. So S3d will happily continue to upload until the backend becomes available again and then pin all files.

Closes https://github.com/SiaFoundation/s3d/issues/189


sia/persist/sqlite/migrations.go
```diff
@@ -24,4 +24,37 @@ var migrations = []func(tx *txn, log *zap.Logger) error{
...
+    filename TEXT,
+    sia_object_id BLOB,
+    sia_object BLOB,
+    CHECK ((sia_object_id IS NULL AND sia_object IS NULL) OR (sia_object_id IS NOT NULL AND sia_object IS NOT NULL)),
+    CHECK ((size = 0 AND filename IS NULL AND sia_object_id IS NULL) OR (size > 0 AND (filename IS NOT NULL OR sia_object_id IS NOT NULL))),
+    PRIMARY KEY (bucket_id, name)
+) WITHOUT ROWID;
+INSERT INTO objects_new (bucket_id, name, content_md5, metadata, size, parts_count, updated_at, filename, sia_object_id, sia_object)
+    SELECT bucket_id, name, content_md5, metadata, size, parts_count, updated_at, filename, sia_object_id, sia_object FROM objects;
+DROP TABLE objects;
```
chris124567, line 49: `object_parts` references `objects` with `ON DELETE CASCADE` so we will lose that data in this migration

sia/pin.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	if err := s.sdk.PinObject(ctx, sdkObj); err != nil {
+		s.logger.Warn("failed to pin object",
+			zap.Stringer("siaObjectID", &id),
+			zap.Error(err))
+		return s.reschedulePin(id)
+	}
+
+	orphans, err := s.store.MarkObjectPinned(id)
```
peterjan, line 164: This might return `objects.ErrObjectNotFound` right? If that happens this error bubbles up and causes a pin backoff, delaying all other pending pins and blocking the queue. Concurrent call to delete might cause this I think.

sia/persist/sqlite/objects.go
```diff
@@ -557,7 +746,16 @@ func insertOrphan(tx *txn, objectID types.Hash256) error {
...
 	}
-	_, err := tx.Exec("INSERT OR IGNORE INTO orphaned_objects (sia_object_id) VALUES ($1)", sqlHash256(objectID))
+	res, err := tx.Exec(`DELETE FROM unpinned_objects WHERE sia_object_id = $1`, sqlHash256(objectID))
+	if err != nil {
+		return err
+	}
+	if n, err := res.RowsAffected(); err != nil {
+		return err
+	} else if n > 0 {
+		return nil // was never pinned; Sia TTL will expire it
```
peterjan: Hm, this seems dangerous, because inserting an entry into `orphaned_objects` is really the only mechanism that cleans up pinned objects, so if we skip that step an object might be pinned forever even though it's no longer referenced. The `unpinned_objects` gets deleted by `MarkObjectPinned` but it's possible for the delete to trigger between the SDK pin call and the mark object pinned call...

peterjan: If confirmed the fix might be to check for `ObjectNotFound` in `MarkObjectPinned` and orphan the pinned object. I think that's safe because the only reason for an object not to be there is because it got deleted... orphan loop can pick it up later

peterjan (changes requested):

peterjan (dismissed):

chris124567 (approved):

PR #197, opened 2026-06-09
v0.1.0

ChrisSchinnerl:
Quick cleanup PR before tagging v0.1.0.

- Reduces some log spam on the info level
- Removes the pre-release changelog files
- Removes warning banner
- Adds knope workflow

### DIFF
--- a/sia/upload.go
+++ b/sia/upload.go
@@ -196,10 +196,9 @@ func (s *Sia) uploadObjects(ctx context.Context) { //nolint:revive
 	if len(groups) == 0 {
 		s.logger.Debug("not enough objects for upload")
 		return
-	} else {
-		s.logger.Info("found enough objects for upload",
-			zap.Int("groups", len(groups)))
 	}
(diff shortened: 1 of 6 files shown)


sia/upload.go
```diff
@@ -194,8 +194,11 @@ func (s *Sia) uploadObjects(ctx context.Context) { //nolint:revive
 	// fetch and prepare objects for upload
 	groups := s.prepareUploads()
 	if len(groups) == 0 {
-		s.logger.Debug("upload loop tick")
+		s.logger.Debug("not enough objects for upload")
 		return
+	} else {
+		s.logger.Info("found enough objects for upload",
```
n8mgr: I think this is the definition of log spam in the info level

ChrisSchinnerl: One line per batch of uploads? Before it was 1 line per minute regardless of whether we were uploading anything 😅

n8mgr (approved):

Issue #202, opened 2026-06-10
Backups


ChrisSchinnerl:
Tracking issue for backup support in S3d. Right now s3d requires manually backing sqlite which is inconvenient and makes it annoying to use s3d as a backup source.

To fix this we can do the following

- Add an endpoint to create a sqlite backup (copy hostd behavior) 
- Add an endpoint to flush objects that haven't been uploaded yet regardless of padding
- Add CLI cmds to flush files and create, list and delete backups 
- Add a recovery command which syncs all objects, lists the snapshot ones, let's the user select one of them and then creates a database from it
- Update the README to explain how it all works and what the implications are

This is quite a bit of code and should be done in multiple PRs. e.g. creating the backup can be one, the flush endpoint can be one, the commands and finally the recovery command that restore the database. 

NOTE: One thing we need to be careful about is making sure that orphaning doesn't break past backups. e.g. if we create a backup and then delete a file, restoring should result in the deleted file still being pinned. So we need to 1. not orphan an object if it is linked to a snapshot and 2. potentially insert orphans when we delete a snapshot

NOTE2: Restoring a snapshot that isn't the latest one will also require careful consideration since snapshots are not independent. e.g. restoring snapshot 1 leaves s3d without knowledge of snapshot 2 and all of its pinned objects. Meaning that in the worst case using snapshot 1 corrupts snapshot 2 and in the best case we collect garbage data.

PR #209, opened 2026-06-16
Add flush endpoint

chris124567:
Re: #202


sia/upload.go
```diff
@@ -191,29 +193,44 @@ func (s *Sia) uploadLoop(ctx context.Context) {
...
+// any pending object remains buffered locally.
+func (s *Sia) FlushObjects(ctx context.Context) error {
+	s.uploadObjects(ctx, true)
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	stats, err := s.store.UploadStats()
+	if err != nil {
+		return fmt.Errorf("failed to check pending objects after flush: %w", err)
+	} else if stats.PendingObjects > 0 {
```
ChrisSchinnerl: This should be fine. It's highly unlikely that no objects get uploaded on a busy s3d instance while the flushing is happening. All that matters for the backup is that all objects that were pending before calling flush were uploaded. Objects uploaded during the flush are just not covered by the backup. That is something we need to be aware of when implementing recovery.

ChrisSchinnerl (changes requested):

sia/upload.go
```diff
@@ -191,14 +199,34 @@ func (s *Sia) uploadLoop(ctx context.Context) {
...
 		}
 	}
 }
 
-func (s *Sia) uploadObjects(ctx context.Context) { //nolint:revive
+// FlushObjects uploads every pending object to Sia regardless of padding,
+// rather than waiting for the background loop to batch them into efficiently
+// packed slabs, and pins the uploaded objects so their local backup files are
+// removed. It blocks until the uploads and pins complete.
+func (s *Sia) FlushObjects(ctx context.Context) error {
```
peterjan, line 213: This really only fails if pinning fails or if the context gets cancelled. Wouldn't it be better if `FlushObjects` tries to upload objects until there are no more pending? Or at the very least error out if uploading fails. It's somewhat of a false sense of security to have a flush method that you likely run before you backup, to later realise the flush didn't really flush anything. I'll rope @ChrisSchinnerl in on this to get his opinion about this as well before making any changes to the PR.

peterjan (changes requested):

sia/upload.go
chris124567: This was discussed a little bit in https://github.com/SiaFoundation/s3d/pull/209#discussion_r3427054741... I guess we can continue that here

peterjan: Oh I'm sorry, I didn't catch that. I don't think `Flush` should run in a loop per se, I do think we should surface an error though if we failed to flush the batch of pending objects at the time of calling flush. I don't feel too strongly though, I just think `FlushObjects` swallows a lot of errors is all.

ChrisSchinnerl: Yeah a failing flush should probably return an error if either uploading or pinning fails.

chris124567: OK added some code to get the error(s) out of `uploadObjects`

sia/upload.go
```diff
@@ -191,17 +199,40 @@ func (s *Sia) uploadLoop(ctx context.Context) {
...
+	if err := s.uploadObjects(ctx, true); err != nil {
+		return fmt.Errorf("failed to upload objects: %w", err)
+	} else if err := ctx.Err(); err != nil {
+		return err
+	} else if err := s.performObjectPinning(ctx); err != nil {
+		// don't wait for the pinning loop
+		return fmt.Errorf("failed to pin objects after flush: %w", err)
+	}
+	return ctx.Err()
```
peterjan: return nil here, we successfully uploaded and pinned the objects

peterjan (dismissed):

peterjan (dismissed):

sia/upload.go
```diff
@@ -191,17 +199,40 @@ func (s *Sia) uploadLoop(ctx context.Context) {
...
+// FlushObjects uploads every pending object to Sia regardless of padding,
+// rather than waiting for the background loop to batch them into efficiently
+// packed slabs, and pins the uploaded objects so their local backup files are
+// removed. It blocks until the uploads and pins complete.
+func (s *Sia) FlushObjects(ctx context.Context) error {
+	if err := s.uploadObjects(ctx, true); err != nil {
+		return fmt.Errorf("failed to upload objects: %w", err)
+	} else if err := ctx.Err(); err != nil {
+		return err
+	} else if err := s.performObjectPinning(ctx); err != nil {
```
ChrisSchinnerl, line 218: `performObjectPinning` currently doesn't guard against the race that `uploadObjects` avoids with the dedicated mutex. It's possible that the loop and flush both fetch the same objects to pin and then produce errors.

sia/upload.go
```diff
@@ -191,17 +199,40 @@ func (s *Sia) uploadLoop(ctx context.Context) {
...
+		// don't wait for the pinning loop
```
ChrisSchinnerl: slightly confusing comment imo

sia/upload.go
```diff
@@ -231,6 +263,7 @@ func (s *Sia) uploadObjects(ctx context.Context) { //nolint:revive
 				err := s.uploadObjectGroup(ctx, g)
 				if err != nil {
 					s.logger.Error("failed to upload object group", zap.Error(err))
+					errs[i] = errors.Join(errs[i], err)
```
ChrisSchinnerl, line 263: This might not be sufficient. uploadObjectGroup doesn't fail on every error. Some are just logged. So the flush might return no error when it really did fail to upload some objects.

ChrisSchinnerl (changes requested):

sia/upload.go
```diff
@@ -402,5 +405,5 @@ func (s *Sia) uploadObjectGroup(ctx context.Context, group uploadGroup) error {
 			zap.String("name", uploadObj.Name))
 		queued++
 	}
-	return nil
+	return errors.Join(errs...)
```
ChrisSchinnerl, line 407: Every error in `errs` will now be logged twice. Once in here and once compressed to a joined error in `uploadObjects`.

ChrisSchinnerl (changes requested):
Apart from the double locking it looks fine now. I'd double change the other places where we log as well now to make sure it's still sane with the joined errors.

peterjan (approved):

ChrisSchinnerl (approved):

Issue #201, opened 2026-06-10
Add support for versioning


ChrisSchinnerl:
Tracking issue for S3 versioning support.

PR #214, opened 2026-06-22
Remainder of versioning

chris124567:
Re: #201

`git diff christopher/versioning-2 christopher/versioning-3` produces no output.

### DIFF
--- a/s3/messages.go
+++ b/s3/messages.go
@@ -120,6 +120,7 @@ type (
 
 	// ListObjectsV1Result is the response to a ListObjects (v1) request.
 	ListObjectsV1Result struct {
+		XMLName xml.Name `xml:"ListBucketResult"`
 		ListObjectsResultBase
 
 		// Indicates where in the bucket listing begins. Echoed from the
@@ -134,6 +135,7 @@ type (
 
 	// ListObjectsV2Result is the response to a ListObjectsV2 request.
 	ListObjectsV2Result struct {
+		XMLName xml.Name `xml:"ListBucketResult"`
 		ListObjectsResultBase
 
 		// If ContinuationToken was sent with the request, it is included in the
@@ -153,7 +155,8 @@ type (
 		StartAfter string `xml:"StartAfter,omitempty"`
 	}
 
-	// Version represents a version of an S3 object.
+	// Version represents a version of an S3 object in a ListObjectVersions
+	// response.
 	Version struct {
 		XMLName      xml.Name    `xml:"Version"`
 		Key          string      `xml:"Key"`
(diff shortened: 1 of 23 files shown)


s3/messages.go
```diff
@@ -169,52 +172,52 @@ type (
...
-		Versions []Version
+		NextVersionIDMarker string `xml:"NextVersionIdMarker,omitempty"`
 
-		// prefixes maintains an index of prefixes that have already been seen.
-		// This is a convenience for backend implementers like s3bolt and s3mem,
-		// which operate on a full, flat list of keys.
-		prefixes map[string]bool
+		// Versions holds the Version and DeleteMarker elements interleaved in
+		// response order: key ascending, then newest version first.
+		Versions []VersionListEntry
```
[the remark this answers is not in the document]
chris124567: Each of the concrete types have XMLName fields so this is not an issue

chris124567:
Test failure is some weird tenki issue....

tox/tox_test.go
```diff
@@ -157,11 +157,20 @@ func TestS3(t *testing.T) {
 		runTox(t, confPath, testsDir,
 			"s3tests/functional/test_s3.py",
 			"-m", "not s3d_not_implemented and not s3d_not_supported and not bucket_logging and not encryption and not sse_s3 and not bucket_encryption and not lifecycle_transition and not tagging and not bucket_policy and not conditional_write and not object_ownership and not checksum and not cloud_transition and not cloud_restore and not iam_user and not iam_account and not delete_marker",
```
peterjan: Can you verify whether `and not delete_marker` is correct here? 

I think that covers these four of which we want to enable a couple now that we have versioning. 
- test_delete_marker_nonversioned
- test_delete_marker_versioned
- test_delete_marker_suspended
- test_delete_marker_expiration

I think it should be

```
runTox(t, confPath, testsDir,
			"s3tests/functional/test_s3.py",
			"-m", "not s3d_not_implemented and not s3d_not_supported and not bucket_logging and not encryption and not sse_s3 and not bucket_encryption and not lifecycle_transition and not tagging and not bucket_policy and not conditional_write and not object_ownership and not checksum and not cloud_transition and not cloud_restore and not iam_user and not iam_account",
			// the lifecycle exclusions drop tests for unimplemented features:
			// tag and object-size filters, noncurrent-version and delete-marker
			// actions, and transitions. lifecycle_expiration_versioned is one of
			// these despite its name: it relies on object tagging. Plain
			// expiration on a versioned bucket is supported and runs.
			//
			// delete_marker_nonversioned expects a HEAD of a deleted key in an
			// unversioned bucket to return x-amz-delete-marker: false. s3d removes
			// the object and returns NoSuchKey without that header.
			// delete_marker_expiration needs the ExpiredObjectDeleteMarker and
			// NoncurrentVersionExpiration lifecycle actions, which are unsupported.
			//
			// bucket_list_return_data_versioning is excluded because it calls
			// get_object_acl, which s3d does not implement. The test name has no
			// _acl suffix, so the marker and keyword filters above do not catch it.
			//
			// bucket_recreate_not_overriding expects re-creating an owned bucket
			// to succeed idempotently, which is the legacy-region behavior; s3d
			// follows the default region and returns BucketAlreadyOwnedByYou.
			"-k", "not _acl and not post_object and not _torrent and not cors and not object_lock and not retention and not legal_hold and not notification and not replication and not website and not _select and not lifecycle_expiration_tags and not lifecycle_expiration_versioned and not lifecycle_expiration_size and not tags_head and not noncur and not deletemarker and not delete_marker_nonversioned and not delete_marker_expiration and not lifecycle_set_filter and not lifecycle_set_empty_filter and not lifecycle_transition_set_invalid_date and not bucket_list_return_data_versioning and not bucket_recreate_not_overriding",
		)
```
but please verify

peterjan (changes requested):
While going over the S3 test suite, I found some tests we'd previously switched off that this versioning work now covers:

```
- test_object_copy_versioned_bucket
- test_object_copy_versioned_url_encoding
- test_multipart_copy_versioned
```

tox/tox_test.go
chris124567: `delete_marker_nonversioned` - deletes a key in an unversioned bucket, then HEADs it and asserts that the 404 response has x-amz-delete-marker: false.  That shouldn't happen according to the [aws spec](https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeleteMarker.html):

> The x-amz-delete-marker: true response header tells you that the object accessed was a delete marker. This response header never returns false, because when the value is false, the current or specified version of the object is not a delete marker.


`delete_marker_expiration` - requires `NoncurrentVersionExpiration` which we haven't implemented yet

I enabled the others


Regarding:
- test_object_copy_versioned_bucket
- test_object_copy_versioned_url_encoding
- test_multipart_copy_versioned

I will do a followup in https://github.com/SiaFoundation/s3-tests

ChrisSchinnerl (approved):
Looking good. I would indeed follow up with a change to the s3-tests since we can probably move tests that we ignore explicitly now into our custom s3d tags.

Tried this with Cyberduck, rclone and Nextcloud and seems to work as expected.

peterjan (approved):

PR #215, opened 2026-06-24
Add snapshots

peterjan:
This PR introduces snapshots. A snapshot is a backup of the sqlite database that gets gzipped and uploaded to the Sia network as a tagged, pinned object, so the full database can be retrieved and restored to its state at the time the snapshot was taken. 

References https://github.com/SiaFoundation/s3d/issues/202


sia/persist/sqlite/snapshots.go
```diff
@@ -0,0 +1,85 @@
...
+		_, err := tx.Exec(`
+			INSERT INTO snapshot_objects (snapshot_id, sia_object_id)
+			SELECT DISTINCT $1, sia_object_id FROM objects WHERE sia_object_id IS NOT NULL`, snapshotID)
+		return err
+	})
+	if err != nil {
+		return err
+	}
+
+	if err := s.Backup(ctx, destPath); err != nil {
```
peterjan: I'm expecting some discussion here... I considered some alternatives but eventually landed on inserting the snapshot first, then backup up. It makes the backup self contained, e.g. it contains the `snapshot_objects`, that should come in handy during recovery. 

I think it makes sense to register the snapshot first, then take the backup. Worst case you are over-protected in the sense that you have a snapshot without sql backup. If you were to do it the other way around you might end up with a sql backup without accompanying snapshot, leaving the objects unprotected as they're not referenced by a snapshot.

sia/persist/sqlite/snapshots.go
```diff
@@ -0,0 +1,97 @@
...
+func (s *Store) CreateSnapshot(ctx context.Context, destPath string) error {
+	if destPath == "" {
+		return errors.New("empty destination path")
+	} else if _, err := os.Stat(destPath); err == nil {
+		return errors.New("destination file already exists")
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("failed to stat destination file: %w", err)
+	}
+
+	// hold the single connection across the snapshot insert and the backup so
```
ChrisSchinnerl: My problem with this approach is that we are potentially inserting thousands or millions of rows during a snapshot. Blocking the database and making s3d pretty much unusable. While the SQLite backup itself could be performed on a live system without interruptions.

How about instead of adding a link for every object we use a more conservative approach and just compare timestamps. If an object in our database has a timestamp smaller than the Min(snapshot.created_at) we assume it might be part of the snapshot and we don't orphan it. 

Snapshots are not an accurate science to begin with due to the flushing and all and keeping too many objects pinned rather than too few is definitely better.

sia/persist/sqlite/snapshots.go
```diff
@@ -0,0 +1,97 @@
...
+	})
+	return
+}
+
+// DeleteSnapshot removes a snapshot and its object references from the store.
+// Objects no longer referenced by any snapshot or live object become eligible
+// for unpinning on the next orphan loop.
+func (s *Store) DeleteSnapshot(snapshotID int64) error {
+	return s.transaction(func(tx *txn) error {
+		_, err := tx.Exec("DELETE FROM snapshots WHERE id = $1", snapshotID)
```
chris124567: Check RowsAffected so we can return 404 in #219

sia/persist/sqlite/backup.go
```diff
@@ -39,7 +39,21 @@ func withSQLiteConn(conn *sql.Conn, fn func(*sqlite3.SQLiteConn) error) error {
...
 // always consistent.
-func (s *Store) Backup(ctx context.Context, destPath string) (err error) {
+func (s *Store) Backup(ctx context.Context, destPath string) error {
+	// initialize the source conn
+	srcConn, err := sqlConn(ctx, s.db)
+	if err != nil {
+		return fmt.Errorf("failed to create source connection: %w", err)
+	}
+	defer srcConn.Close()
+	return execBackup(ctx, srcConn, destPath)
```
chris124567: do we still need this function if we have CreateSnapshot?

peterjan: not really but it's probably fine to keep around? It has a nice little isolated test and we might eventually expose a raw backup command like you did that doesn't care about the snapshot aspects of things. I'm fine to delete it though if we all deem it's silly

sia/persist/sqlite/snapshots.go
peterjan: Hm ok I see you on the inserts, that's definitely not ideal but I figured it was worth the trade-off because we would be able to take a precise snapshot and recover from it correctly. 

I don't really like your alternative suggestion to be honest. We would indeed over-pin, but not just a little, objects would essentially stay around forever. We also don't really have a good candidate for the timestamp currently. We only have updated_at, which we update when an object is overwritten, and we keep it in seconds. That would make objects uploaded at the time of snapshot potentially fall out of it.

What about this though: we keep a "generation counter" and both snapshots and orphaned objects reference it. When you create a snapshot you get the value of the counter at that point in time. When an object is orphaned we assign it the current counter value and the orphan loop only unpins an object whose stamp is below the lowest surviving snapshot... That makes it so if you delete a snapshot, you raise the floor and you release any orphaned objects it may be holding.

I implemented that for the time being because it has a lot of upsides and little to no downsides I think.

sia/persist/sqlite/snapshots.go
```diff
@@ -0,0 +1,107 @@
...
+	} else if _, err := os.Stat(destPath); err == nil {
+		return errors.New("destination file already exists")
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("failed to stat destination file: %w", err)
+	}
+
+	// hold the single connection across the generation bump and the backup so
+	// no writer can orphan an object the backup references before it is stamped
+	// with this snapshot's generation
+	conn, err := sqlConn(ctx, s.db)
```
ChrisSchinnerl: Isn't this effectively an exclusive lock on the whole database? I think the new generation strategy should allow us to avoid that altogether by bumping the generation + inserting a snapshot atomically and then performing the backup during regular operation.

peterjan:
(still working on exploring the trade-offs between backup API and vacuum into)

peterjan:
@ChrisSchinnerl 

I explored several strategies and ended up reverting it to the simplest approach. The main reason why I think it's not trivial, is because of `SetMaxOpenConns`. We set that to one to avoid "database is locked" errors, but this also means that concurrent writers will be blocked for the duration of the backup. I tested and benchmarked this and it does effectively block writers. It's effectively a semaphore right so that makes sense. I think for our other apps that's probably fine but I feel this would be a big downside for s3d. In practice it's likely not going to be an issue unless you have a (really) really big node. 

An alternative approach I considered is to use `VACUUM INTO`. The upside with vacuum is that it compacts the database, so you get the smallest possible backup, and that it doesn't block concurrent writing processes what so ever. The big downside however is that it writes all changes to a WAL during the backup. Considering that we have disk size limits already I figured we probably don't want that. In my testing it generated a 50GB WAL backing up a 40GB database, but that was with 10 write processes hammering it for the whole duration of the backup, still.. not ideal.

The most correct approach would be to gate our single connection and use a transaction queue. All transactions flow through that as well as the backup and essentially what it does is it interleaves "regular operations" with backup steps properly. In a way that you are also guaranteed that the backup process makes progress at fixed intervals and doesn't have to fight for the connection. I was implementing that but it felt over-engineered and figured I'd keep it super simple and propose the approach first. I would say it's 200 lines or something, you'd have a txQueue and a txJob and if there's an active backup you step every couple of jobs or after a fixed amount of time... It comes with some footguns and needless overhead because 99.99% of the time you are not backing up.

We could also forget about FIFO entirely and just rely that the backup will eventually finish because after releasing the connection, it'll fight for the connection again with other concurrent write processes and eventually get it and make progress. In practice that  might work really well. I wrote that out and tested it and in my benchmark the longest wait was 30s, which I found unacceptable.. At the same time though you eventually finish and you don't block writers during the backup process.

In short:
- VACUUM is not ideal because the WAL it produces but it allows concurrent writes, which is really nice
- single conn backup blocks other writers for the duration of the backup, takes ~4m for a 40GB database, but realistically it's very fast because s3d databases don't grow that big
- multi conn backup potentially never finishes because writes make it restart
- single conn backup with interleaved writes is the cleanest approach


edit:

I have a bunch of benchmark output but wanted to type it out instead.

<details>
<summary>SQLite backup strategy comparison</summary>

| Strategy | Completed under load | Duration | Restarts | Steps | Source WAL growth | Backup size | Copy consistent |
|---|:---:|---:|---:|---:|---:|---:|:---:|
| **VACUUM INTO** (dedicated connection) | Yes | **2m 26s** | N/A | 1 statement | **55.3 GB** | 37.0 GB | Yes, 150000/150000 |
| Single-connection backup API | Yes | 3m 35s | 0 | 90,301 | ~0 | 37.2 GB | Yes, 150000/150000 |
| Multi-connection backup API | No, timed out | 5m 00s | **128,884** | 2.48M | ~40 MB | Not produced | No, livelock |
| Interleaved backup API, same connection per step | Yes | 4m 42s | 0 | 90,339 | ~0 | 37.2 GB | Yes, 150000/150000 |

| Strategy | Writes completed | Throughput | Mean latency | Max latency | % < 1 ms | % < 10 ms |
|---|---:|---:|---:|---:|---:|---:|
| **VACUUM INTO** | 13.2M | ~90K/s | 114 µs | 5.60 s | 99.87% | 99.99% |
| Single-connection backup API | 0, fully blocked | 0 | N/A | N/A | N/A | N/A |
| Multi-connection backup API | 16.0M | ~53K/s | 188 µs | 2.46 s | 98.83% | 99.90% |
| Interleaved backup API, same connection per step | 0.9M | ~3.2K/s | 3.11 ms | **30.03 s** | 53.2% | 92.9% |

</details>

edit 2:

One thing to note is that whatever strategy we choose, we should probably do the following items because it's little cost and real benefit:

- integrity_check + fsync + atomic rename publish  for crash-safety
- DELETE-mode destination, we currently don't do that
- write to a temp path, validate, then rename into place
- VACUUM INTO the tmp file into the desintation path to get the compacting

sia/persist/sqlite/snapshots.go
```diff
@@ -0,0 +1,95 @@
...
+		return tx.QueryRow(`
+			INSERT INTO snapshots (created_at, path, gen, object_count)
+			VALUES ($1, $2, $3, (SELECT stat_value FROM stats WHERE stat = $4))
+			RETURNING id`, sqlTime(time.Now()), destPath, gen, statUploadedObjects).Scan(&snapshotID)
+	}); err != nil {
+		return fmt.Errorf("failed to create snapshot: %w", err)
+	}
+
+	// proceed to backup the database, rolling back the snapshot if it fails
+	if err := s.Backup(ctx, destPath); err != nil {
```
ChrisSchinnerl: The logic for creating a snapshot looks pretty good now but one thing missing between this PR and the command for creating the snapshot is the upload to the Sia network.

The goal is that you can install s3d after your machine blows up and then perform a sync for backups using the corresponding command. That sync isn't just listing snapshots from the database but has to fetch snapshots by listing all objects from the indexer.

So for this PR that probably means the schema doesn't even need a path since the snapshot will only exist temporarily and then gets uploaded to the network. Instead we can store the object id on the Sia network. 
That also means the backup endpoint doesn't really need a path. We can just pick a temporary one.

ChrisSchinnerl (changes requested):

sia/persist/sqlite/snapshots.go
peterjan: Ok, I ended up adding a snapshot route alongside the SQLite backups considering that's already in master and we might want to expose the functionality later through a backup command that outputs a backup of the SQLite database. I figured the snapshot functionality could co-exist alongside the backup functionality...

ChrisSchinnerl: Let's stick to a single endpoint for now. We really don't want users to perform manual backups or recover manually by copying over a database. That is just a guarantee for inconsistencies.

sia/persist/sqlite/init.sql
```diff
@@ -130,6 +139,7 @@ CREATE TABLE global_settings (
 	indexer_url TEXT,
 	last_sync_at INTEGER NOT NULL DEFAULT 0,
 	last_sync_key BLOB NOT NULL DEFAULT X'0000000000000000000000000000000000000000000000000000000000000000',
+	snapshot_generation INTEGER NOT NULL DEFAULT 0,
```
ChrisSchinnerl: (nit) only field where we spell generation out. In other tables we just call it gen.

sia/persist/sqlite/snapshots.go
```diff
@@ -0,0 +1,84 @@
...
+		return tx.QueryRow(`
+			INSERT INTO snapshots (created_at, gen, object_count)
+			VALUES ($1, $2, (SELECT stat_value FROM stats WHERE stat = $3))
+			RETURNING id, created_at, object_count`, sqlTime(time.Now()), gen, statUploadedObjects).Scan(&snap.ID, (*sqlTime)(&snap.CreatedAt), &snap.ObjectCount)
+	})
+	return
+}
+
+// SetSnapshotSiaObject records the Sia object ID for an uploaded snapshot.
+func (s *Store) SetSnapshotSiaObject(id int64, objectID types.Hash256) error {
```
ChrisSchinnerl: `MarkSnapshotPinned` for consistency?

sia/sia.go
ChrisSchinnerl, line 1: We got a `syncMetadata` method somewhere in this file which we should probably also update to call `store.DeleteSnapshot` when we receive an event for a deleted snapshot.

We usually want to delete the snapshot object from Sia first and then delete it locally to make sure we don't end up pruning orphans that belong to a snapshot. But there is a race between those actions where the snapshot might be unpinned from the indexer but still exist on disk.

Handling deletion events can close that gap since even if deleting the snapshot from the database gets interrupted, we will eventually notice the deletion event and can then remove it.

ChrisSchinnerl (changes requested):

chris124567 (commented):
Nothing to add to Chris's comments

sia/persist/sqlite/snapshots.go
peterjan: Ok well then I got rid of `/system/sqlite/backup` in favour of `POST /snapshots`. I think that makes sense but I can see why you would disagree. I also gzip the backup before uploading, vup does that too and it does save a lot of space but more importantly also faster to upload the snapshot to the network.

sia/sia.go
peterjan: Didn't consider this but makes sense

sia/sia.go
```diff
@@ -271,6 +283,29 @@ func New(ctx context.Context, sdk SDK, store Store, directory string, opts ...Op
 		sia.logger.Info("removed orphaned uploads", zap.Int("removed", deleted))
 	}
 
+	// remove snapshot backups left behind by a crash
+	tmpFiles, err := filepath.Glob(filepath.Join(sia.directory, "snapshot-*.tmp"))
```
ChrisSchinnerl: (nit) it's a bit awkward that snapshots pollute the root data dir of s3d. Would be simpler if it just created a `tmp` dir which we delete all files from on startup.

sia/sia.go
```diff
@@ -302,16 +337,105 @@ func (s *Sia) Close() error {
...
+	}()
+	// closing the read side unblocks the compressor if the upload aborts early
+	defer pr.Close()
+
+	obj := sdk.NewEmptyObject()
+	obj.UpdateMetadata(meta)
+	if err := s.sdk.Upload(ctx, &obj, pr); err != nil {
+		return s3.Snapshot{}, fmt.Errorf("failed to upload snapshot: %w", err)
+	}
+	pinned = obj.ID()
```
ChrisSchinnerl: why not set `pinned` after `PinObject`? Then you know if it is set, it is actually pinned.

Then you should also be able to get rid of the redundant cleanup code since the sync loop will check a newly pinned snapshot against the database and if it doesn't exist unpin it.

sia/sia.go
```diff
@@ -489,18 +613,37 @@ func (s *Sia) syncMetadata(ctx context.Context) { //nolint:revive
...
+		// consume events in order, stopping at the first snapshot object that
+		// can't be reconciled yet so it is retried on the next sync
 		var batch []objects.SiaObject
+		processed := 0
 		for _, ev := range events {
 			if ev.Deleted {
-				s.logger.Debug("skipping deleted object event", zap.Stringer("objectID", &ev.Key))
+				// drop snapshots whose backup object was deleted on the
+				// indexer so they stop withholding orphans
+				if n, err := s.store.DeleteSnapshotsBySiaObject(ev.Key); err != nil {
```
ChrisSchinnerl: This is called even without knowing whether the object is a snapshot. Seems like a query we can avoid.

sia/sia.go
```diff
@@ -489,18 +613,37 @@ func (s *Sia) syncMetadata(ctx context.Context) { //nolint:revive
...
+					s.logger.Info("deleted snapshot for deleted object", zap.Stringer("objectID", &ev.Key))
+				}
+				processed++
 				continue
 			} else if ev.Object == nil {
 				s.logger.Warn("skipping event with nil object", zap.Stringer("objectID", &ev.Key))
+				processed++
+				continue
+			} else if isSnapshotObject(ev.Object) {
+				if !s.confirmSnapshotObject(ctx, ev.Key) {
```
ChrisSchinnerl: `confirmSnapshotObject` is a bit of an odd name for something that potentially unpins. Especially if it can lead to the loop being broken out of.

sia/sia.go
```diff
@@ -512,16 +655,56 @@ func (s *Sia) syncMetadata(ctx context.Context) { //nolint:revive
...
+// reconciliation is deferred until no upload is running.
+func (s *Sia) confirmSnapshotObject(ctx context.Context, objectID types.Hash256) bool {
+	known, inFlight, err := s.store.CheckSnapshotObject(objectID)
+	if err != nil {
+		s.logger.Error("failed to check snapshot object", zap.Stringer("objectID", &objectID), zap.Error(err))
+		return false
+	} else if known {
+		return true
+	} else if inFlight {
+		return false
```
ChrisSchinnerl, line 961: The doctoring says that a snapshot that has no local record due to crashing after pinning but before setting the object id in the database gets unpinned. However, in that case `inFlight` is `true` which means not only that it doesn't get unpinned but also that the caller will break out of the sync loop. Never syncing again.

Considering that `DeleteIncompleteSnapshots` is only called on startup, this might be stuck forever.

ChrisSchinnerl (changes requested):

sia/sia.go
peterjan: Yeah I was a bit upset when I found out too but the only other alternative I see is keeping snapshot IDs in memory, which is probably not worth it. We can't know/check whether it's a snapshot from the event because deleted events don't have the metadata. The snapshots table is tiny and a delete should be a relatively rare event type, not sure if it's worth getting fancy over?

sia/sia.go
peterjan: A crash means it restarted though right, in which case the `DeleteIncompleteSnapshots` will have cleared the row in which case `inFlight` is false.

edit: I'm working on testing restoring a backup and I found out I have to update the sync loop's behaviour anyway. Instead of unpinning we have to adopt the snapshot. That's because the backup actually doesn't contain the `sia_object_id` - so it would actually be treated as an incomplete snapshot

sia/sia.go
peterjan: I agree, renamed it.

sia/sia.go
peterjan: this is no longer relevant after recent changes

peterjan:
going to push one more commit that adds the `synced` concept that gates orphan processing to ensure we don't mess up while snapshots are being actively restored

chris124567:
@peterjan vibed this up.  Seems like the fix is somewhat difficult...

```
diff --git a/sia/persist/sqlite/snapshots_test.go b/sia/persist/sqlite/snapshots_test.go
index ad35449..e4e37ac 100644
--- a/sia/persist/sqlite/snapshots_test.go
+++ b/sia/persist/sqlite/snapshots_test.go
@@ -227,3 +227,52 @@ func TestSnapshots(t *testing.T) {
 		t.Fatal("unexpected", orphans)
 	}
 }
+
+// TestSnapshotDoesNotRetainFutureObject demonstrates that a snapshot must not
+// retain an object that was created after the snapshot was taken. The current
+// generation-only orphan guard fails this test because both the old snapshot
+// and the later object share the same generation.
+func TestSnapshotDoesNotRetainFutureObject(t *testing.T) {
+	const bucket = "test-bucket"
+
+	store := initTestDB(t, zaptest.NewLogger(t))
+	if err := store.CreateBucket(testAccessKeyID, bucket); err != nil {
+		t.Fatal(err)
+	}
+
+	// Take and complete an empty snapshot.
+	snap, _, err := store.CreateSnapshot()
+	if err != nil {
+		t.Fatal(err)
+	} else if snap.ObjectCount != 0 {
+		t.Fatal("expected an empty snapshot, got", snap.ObjectCount, "objects")
+	} else if err := store.MarkSnapshotPinned(snap.ID, frand.Entropy256()); err != nil {
+		t.Fatal(err)
+	}
+
+	// Upload and pin an object after the snapshot was taken.
+	obj := sdk.Object{}
+	sealed := obj.Seal(types.GeneratePrivateKey())
+	objectID := sealed.ID()
+	md5 := frand.Entropy128()
+	filename := "future-object"
+	if _, _, err := store.PutObject(testAccessKeyID, bucket, "future", md5, nil, 1, &filename); err != nil {
+		t.Fatal(err)
+	} else if err := store.MarkObjectUploaded(bucket, "future", "", md5, sealed, time.Now().Add(time.Hour)); err != nil {
+		t.Fatal(err)
+	} else if _, err := store.MarkObjectPinned(objectID); err != nil {
+		t.Fatal(err)
+	}
+
+	// Since the snapshot predates the object, deleting the object should make
+	// it immediately eligible for unpinning.
+	if _, _, _, err := store.DeleteObject(testAccessKeyID, bucket, s3.ObjectID{Key: "future"}); err != nil {
+		t.Fatal(err)
+	}
+	orphans, err := store.OrphanedObjects(100)
+	if err != nil {
+		t.Fatal(err)
+	} else if len(orphans) != 1 || orphans[0] != objectID {
+		t.Fatalf("snapshot predates object: expected orphan %v to be eligible, got %v", objectID, orphans)
+	}
+}
```

```
$ go test ./sia/persist/sqlite -v -count=1 -run "TestSnapshotDoesNotRetainFutureObject"
=== RUN   TestSnapshotDoesNotRetainFutureObject
    logger.go:146: 2026-07-16T17:07:52.865-0400 DEBUG   sqlite  database initialized    {"sqliteVersion": "3.53.2", "schemaVersion": 9, "path": "/tmp/TestSnapshotDoesNotRetainFutureObject2403215939/001/s3d.sqlite"}
    snapshots_test.go:276: snapshot predates object: expected orphan 0e5751c026e543b2e8ab2eb06099daa1d1e5df47778f7787faab45cdf12fe3a8 to be eligible, got []
--- FAIL: TestSnapshotDoesNotRetainFutureObject (0.01s)
FAIL
FAIL    github.com/SiaFoundation/s3d/sia/persist/sqlite 0.020s
```

peterjan:
> @peterjan vibed this up. Seems like the fix is somewhat difficult...
> 
> ```
> diff --git a/sia/persist/sqlite/snapshots_test.go b/sia/persist/sqlite/snapshots_test.go
> index ad35449..e4e37ac 100644
> ...
> 0e5751c026e543b2e8ab2eb06099daa1d1e5df47778f7787faab45cdf12fe3a8 to be eligible, got []
> --- FAIL: TestSnapshotDoesNotRetainFutureObject (0.01s)
> FAIL
> FAIL    github.com/SiaFoundation/s3d/sia/persist/sqlite 0.020s
> ```

Thank you, I'll look into it. There's many many edges with snapshot recovery, at least this one is retaining the object which is definitely preferable over unpinning an object that we want to keep. But I'll see what the issue is and try and come up with a good fix for it.

peterjan:
> @peterjan vibed this up. Seems like the fix is somewhat difficult...

Hm yeah, I ended up bumping the generation again when a backup finishes and recording it as `gen_completed` on the snapshot. That gives us a window and tells us when we can orphan an object that was created after the snapshot was taken and deleted swiftly after. Extended tests to cover the scenario.

chris124567:
Harder to trigger but here is fix and regression test

```
diff --git a/sia/persist/sqlite/snapshots.go b/sia/persist/sqlite/snapshots.go
index 8dfaea0..98b0eb9 100644
--- a/sia/persist/sqlite/snapshots.go
+++ b/sia/persist/sqlite/snapshots.go
@@ -42,11 +42,13 @@ func (s *Store) CreateSnapshot() (snap s3.Snapshot, gen int64, err error) {
 
 // AdoptSnapshot records a snapshot for a backup object discovered on the
 // network, e.g. after restoring from a backup made by a previous database.
-// The generation counter is raised to at least the adopted generation, and
-// existing orphans with it, since the adopted backup may reference them. As
-// in MarkSnapshotPinned the counter is then bumped once more and recorded as
-// the completion generation. Adopting an object that already has a record
-// returns the existing snapshot.
+// The generation counter is raised to at least the adopted generation.
+// Existing orphans are raised to it even when the counter is already past it:
+// the adopted generation is from a different counter history, so it cannot be
+// ordered against locally recorded orphans and the adopted backup may
+// reference any of them. As in MarkSnapshotPinned the counter is then bumped
+// once more and recorded as the completion generation. Adopting an object
+// that already has a record returns the existing snapshot.
 func (s *Store) AdoptSnapshot(objectID types.Hash256, createdAt time.Time, gen, objectCount int64) (snap s3.Snapshot, err error) {
 	err = s.transaction(func(tx *txn) error {
 		// return the existing record when the object was already adopted
@@ -59,16 +61,10 @@ func (s *Store) AdoptSnapshot(objectID types.Hash256, createdAt time.Time, gen,
 			return err
 		}
 
-		var current int64
-		if err := tx.QueryRow("SELECT snapshot_gen FROM global_settings").Scan(&current); err != nil {
+		if _, err := tx.Exec("UPDATE orphaned_objects SET orphaned_at_gen = $1 WHERE orphaned_at_gen < $1", gen); err != nil {
+			return err
+		} else if _, err := tx.Exec("UPDATE global_settings SET snapshot_gen = $1 WHERE snapshot_gen < $1", gen); err != nil {
 			return err
-		}
-		if gen > current {
-			if _, err := tx.Exec("UPDATE orphaned_objects SET orphaned_at_gen = $1 WHERE orphaned_at_gen < $1", gen); err != nil {
-				return err
-			} else if _, err := tx.Exec("UPDATE global_settings SET snapshot_gen = $1", gen); err != nil {
-				return err
-			}
 		}
 
 		var completed int64
diff --git a/sia/persist/sqlite/snapshots_test.go b/sia/persist/sqlite/snapshots_test.go
index d8ec4ec..1aebe19 100644
--- a/sia/persist/sqlite/snapshots_test.go
+++ b/sia/persist/sqlite/snapshots_test.go
@@ -344,3 +344,90 @@ func TestSnapshots(t *testing.T) {
 		t.Fatal("unexpected", orphans)
 	}
 }
+
+// TestAdoptSnapshotWithholdsExistingOrphans verifies that an adopted snapshot
+// withholds orphans recorded before the adoption even when an earlier adoption
+// advanced the local counter to its generation. This can happen when snapshots
+// from different histories remain on the network after multiple restores.
+func TestAdoptSnapshotWithholdsExistingOrphans(t *testing.T) {
+	const bucket = "test-bucket"
+
+	store := initTestDB(t, zaptest.NewLogger(t))
+
+	if err := store.CreateBucket(testAccessKeyID, bucket); err != nil {
+		t.Fatal(err)
+	}
+
+	// upload and pin an object that exists in the database being restored
+	obj := newTestObject()
+	sealed := obj.Seal(types.GeneratePrivateKey())
+	objID := sealed.ID()
+	md5 := frand.Entropy128()
+	if _, _, err := store.PutObject(testAccessKeyID, bucket, "a", md5, nil, 1, new(string)); err != nil {
+		t.Fatal(err)
+	} else if err := store.MarkObjectUploaded(bucket, "a", "", md5, sealed, time.Now().Add(time.Hour)); err != nil {
+		t.Fatal(err)
+	} else if _, err := store.MarkObjectPinned(objID); err != nil {
+		t.Fatal(err)
+	}
+
+	// model the restored database image. Its own snapshot row was incomplete
+	// when the image was made, so startup removes the row but retains the
+	// generation counter.
+	restored, restoredGen, err := store.CreateSnapshot()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if n, err := store.DeleteIncompleteSnapshots(); err != nil {
+		t.Fatal(err)
+	} else if n != 1 {
+		t.Fatal("unexpected", n)
+	}
+
+	// this history deletes the object before discovering the snapshots still
+	// present on the network
+	if _, _, _, err := store.DeleteObject(testAccessKeyID, bucket, s3.ObjectID{Key: "a"}); err != nil {
+		t.Fatal(err)
+	}
+
+	// adopt the restored image's own snapshot. It withholds the orphan and its
+	// completion bump advances the counter to the next generation.
+	earlier, err := store.AdoptSnapshot(frand.Entropy256(), restored.CreatedAt, restoredGen, restored.ObjectCount)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// another history created a snapshot immediately after restoring the same
+	// image, so its generation is now equal to the counter advanced above. Its
+	// backup may still reference the locally orphaned object.
+	adopted, err := store.AdoptSnapshot(frand.Entropy256(), time.Now(), restoredGen+1, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if orphans, err := store.OrphanedObjects(100); err != nil {
+		t.Fatal(err)
+	} else if len(orphans) != 0 {
+		t.Fatal("unexpected", orphans)
+	}
+
+	// deleting the earlier snapshot must leave the orphan withheld by the later
+	// snapshot
+	if err := store.DeleteSnapshot(earlier.ID); err != nil {
+		t.Fatal(err)
+	}
+	if orphans, err := store.OrphanedObjects(100); err != nil {
+		t.Fatal(err)
+	} else if len(orphans) != 0 {
+		t.Fatal("unexpected", orphans)
+	}
+
+	// deleting the later snapshot releases the orphan
+	if err := store.DeleteSnapshot(adopted.ID); err != nil {
+		t.Fatal(err)
+	}
+	if orphans, err := store.OrphanedObjects(100); err != nil {
+		t.Fatal(err)
+	} else if len(orphans) != 1 || orphans[0] != objID {
+		t.Fatal("unexpected", orphans)
+	}
+}
```

peterjan:
@n8mgr with Chris being OOO I decided to tag you as a reviewer on snapshots as well. Not sure why I didn't do that in the first place, it was not the intention of the PR initially to tackle seed based recovery, almost in-full. It makes sense you have eyes on this as well. (I'm on this + Go SDK parity by the way, just to give you some insights on my focus now that Chris is away).

sia/persist/sqlite/init.sql
```diff
@@ -79,6 +79,7 @@ CREATE TABLE objects (
     updated_at INTEGER NOT NULL,
     filename TEXT, -- name of file for regular uploads or dir for multipart uploads
     sia_object_id BLOB REFERENCES sia_objects(id),
+    created_at_gen INTEGER NOT NULL DEFAULT 0,
```
ChrisSchinnerl: I understand that this is the easier route but it's quite important that these generations are set correctly. So I'd drop the default here and in orphaned_objects and write the more verbose migrations.

sia/persist/sqlite/init.sql
```diff
@@ -135,8 +136,25 @@ CREATE TABLE object_parts (
...
+    orphaned_at_gen INTEGER NOT NULL DEFAULT 0,
+    created_at_gen INTEGER NOT NULL DEFAULT 0 -- creation stamp of the last object row that referenced the id
 );
+CREATE INDEX orphaned_objects_gen_idx ON orphaned_objects(orphaned_at_gen);
+
+CREATE TABLE snapshots (
+    id INTEGER PRIMARY KEY,
+    created_at INTEGER NOT NULL,
+    sia_object_id BLOB,
+    gen INTEGER NOT NULL DEFAULT 0,
```
ChrisSchinnerl: Same here and in `bucket_lifecycle_configurations`. There is on reason this wouldn't be set explicitly right?

sia/persist/sqlite/objects.go
```diff
@@ -669,6 +671,16 @@ func (s *Store) CopyObject(accessKeyID, srcBucket, srcName string, srcVersion s3
 		orphan = res.orphanFile
 		dstVersion := res.dbVersionID
 
+		// the copy inherits the source's creation stamp so their shared
+		// sia_object_id never looks younger than its oldest reference
```
ChrisSchinnerl: This feels like it's in the wrong place. If all objects that share a `sia_object` should have the same creation stamp, I'd move the `created_at_gen` into the `sia_objects` table. There we can just not overwrite it when we upsert.

That should also have a few more side effects like no longer having to pass `createdAtGen` to `insertOrphan` since it can be returned from `deleteSiaObject`.

sia/sia.go
```diff
@@ -486,21 +623,43 @@ func (s *Sia) syncMetadata(ctx context.Context) { //nolint:revive
...
+				}
+				processed++
 				continue
 			} else if ev.Object == nil {
 				s.logger.Warn("skipping event with nil object", zap.Stringer("objectID", &ev.Key))
+				processed++
+				continue
+			} else if meta, ok := snapshotMetadata(ev.Object); ok {
+				if !s.handleSnapshotObject(ev.Key, meta) {
+					break
```
ChrisSchinnerl: Shouldn't this return an error and properly fail? Same for the other ones where we log an error and then just call `SetObjectsCursor`. If for whatever reason there is a bug that prevents adding snapshots we might think we are synced and start processing orphans which should be withheld.

ChrisSchinnerl (changes requested):

sia/sia.go
```diff
@@ -1,7 +1,9 @@
 package sia
```
chris124567, line 1: WWe should do something like this to be more thorough

```
diff --git a/sia/objects/objects.go b/sia/objects/objects.go
index 0d2098f..8080bda 100644
--- a/sia/objects/objects.go
+++ b/sia/objects/objects.go
@@ -2,6 +2,7 @@ package objects
 
 import (
    "errors"
+   "fmt"
    "time"
 
    "go.sia.tech/core/types"
@@ -104,3 +105,25 @@ type SnapshotMetadata struct {
    ObjectCount int64     `json:"objectCount"`
    S3DVersion  string    `json:"s3dVersion"`
 }
+
+// Validate rejects metadata no version of s3d can have written. Unknown
+// encodings and database versions pass so a snapshot from a newer s3d is still
+// recognized. S3DVersion is unchecked, it is empty in unstamped builds.
+func (m SnapshotMetadata) Validate() error {
+   switch {
+   case m.Type != SnapshotType:
+       return fmt.Errorf("unexpected type %q", m.Type)
+   case m.CreatedAt.IsZero():
+       return errors.New("missing creation time")
+   case m.Encoding == "":
+       return errors.New("missing encoding")
+   case m.DBVersion <= 0:
+       return fmt.Errorf("invalid database version %d", m.DBVersion)
+   case m.Generation <= 0:
+       // CreateSnapshot bumps the generation before inserting, so it is never 0
+       return fmt.Errorf("invalid generation %d", m.Generation)
+   case m.ObjectCount < 0:
+       return fmt.Errorf("invalid object count %d", m.ObjectCount)
+   }
+   return nil
+}
diff --git a/sia/objects_test.go b/sia/objects_test.go
index be51adb..d6521ef 100644
--- a/sia/objects_test.go
+++ b/sia/objects_test.go
@@ -1038,29 +1038,36 @@ func TestSyncMetadata(t *testing.T) {
 
    // uploads an object tagged as a snapshot backup to the SDK without
    // recording it locally
-   snapMeta, err := json.Marshal(objects.SnapshotMetadata{Type: objects.SnapshotType})
-   if err != nil {
-       t.Fatal(err)
-   }
-   uploadSnapshotObject := func() sdk.Object {
+   uploadSnapshotObject := func(meta objects.SnapshotMetadata) sdk.Object {
        t.Helper()
+       raw, err := json.Marshal(meta)
+       if err != nil {
+           t.Fatal(err)
+       }
        obj := sdk.NewEmptyObject()
-       obj.UpdateMetadata(snapMeta)
+       obj.UpdateMetadata(raw)
        if err := memSDK.Upload(t.Context(), &obj, bytes.NewReader(frand.Bytes(8))); err != nil {
            t.Fatal(err)
        }
        return obj
    }
+   snapMeta := objects.SnapshotMetadata{
+       Type:       objects.SnapshotType,
+       CreatedAt:  eventTime,
+       DBVersion:  store.DBVersion(),
+       Encoding:   objects.SnapshotEncodingGzip,
+       Generation: 1,
+   }
 
    // record one snapshot object locally, leak another with no local record
-   recorded := uploadSnapshotObject()
+   recorded := uploadSnapshotObject(snapMeta)
    recordedSnap, _, err := store.CreateSnapshot()
    if err != nil {
        t.Fatal(err)
    } else if err := store.MarkSnapshotPinned(recordedSnap.ID, recorded.ID()); err != nil {
        t.Fatal(err)
    }
-   leaked := uploadSnapshotObject()
+   leaked := uploadSnapshotObject(snapMeta)
 
    // the sync leaves both objects pinned and adopts the leaked one, an object
    // without a local record may be a backup made by a previous database
@@ -1084,7 +1091,7 @@ func TestSyncMetadata(t *testing.T) {
 
    // an unknown snapshot object is left alone while an upload is in flight,
    // and the cursor stops before its deferred event
-   blocked := uploadSnapshotObject()
+   blocked := uploadSnapshotObject(snapMeta)
    inflight, _, err := store.CreateSnapshot()
    if err != nil {
        t.Fatal(err)
@@ -1124,6 +1131,27 @@ func TestSyncMetadata(t *testing.T) {
    } else if !cursor.After.Equal(eventTime.Add(5 * time.Second)) {
        t.Fatalf("expected cursor at %v, got %v", eventTime.Add(5*time.Second), cursor.After)
    }
+
+   // a tagged object with metadata s3d can't have written is left pinned
+   // instead of adopted, and the cursor still advances past it
+   malformed := uploadSnapshotObject(objects.SnapshotMetadata{Type: objects.SnapshotType})
+   memSDK.SetEvents([]sdk.ObjectEvent{
+       {Key: malformed.ID(), UpdatedAt: eventTime.Add(6 * time.Second), Object: &malformed},
+   })
+   siaBackend.SyncMetadata(t.Context())
+   if !memSDK.Pinned(malformed.ID()) {
+       t.Fatal("expected malformed snapshot object to stay pinned")
+   }
+   if snapshots, err := store.ListSnapshots(); err != nil {
+       t.Fatal(err)
+   } else if len(snapshots) != 3 {
+       t.Fatal("unexpected", len(snapshots))
+   }
+   if cursor, err := store.ObjectsCursor(); err != nil {
+       t.Fatal(err)
+   } else if !cursor.After.Equal(eventTime.Add(6 * time.Second)) {
+       t.Fatalf("expected cursor at %v, got %v", eventTime.Add(6*time.Second), cursor.After)
+   }
 }
 
 func TestCopyAndDeleteObject(t *testing.T) {
diff --git a/sia/sia.go b/sia/sia.go
index c954f72..7cb27d8 100644
--- a/sia/sia.go
+++ b/sia/sia.go
@@ -706,8 +706,15 @@ func snapshotMetadata(obj *sdk.Object) (objects.SnapshotMetadata, bool) {
 // after restoring from a backup made by a previous database. It reports
 // whether the event was handled. While a snapshot upload is in flight its
 // object may not be recorded yet, so handling is deferred until no upload is
-// running.
+// running. Metadata s3d can't have written is ignored, leaving the object
+// pinned.
 func (s *Sia) handleSnapshotObject(objectID types.Hash256, meta objects.SnapshotMetadata) bool {
+   if err := meta.Validate(); err != nil {
+       // adopting would withhold every orphan the object might reference, and
+       // deferring would stall the cursor and block orphan processing
+       s.logger.Warn("ignoring malformed snapshot object", zap.Stringer("objectID", &objectID), zap.Error(err))
+       return true
+   }
    known, inFlight, err := s.store.CheckSnapshotObject(objectID)
    if err != nil {
        s.logger.Error("failed to check snapshot object", zap.Stringer("objectID", &objectID), zap.Error(err))
```

peterjan: nice catch, applied

sia/persist/sqlite/init.sql
peterjan: This complicated the migration a lot but I guess it's the right thing to do indeed.

sia/persist/sqlite/objects.go
peterjan: Agreed, moved it.

sia/persist/sqlite/init.sql
peterjan: I got rid of all unnecessary defaults. Not sure I understand the `bucket_lifecycle_configurations` reference?

sia/sia.go
peterjan: Ok, made some changes. The cursor part was already covered I think. Every error path breaks before we bump `processed`, so the cursor only advances to the last event we actually handled and the failing one gets retried on the next pass. And `synced` only flips on the branch where the stream comes back empty, so a pass that dies halfway never marks itself synced.

peterjan:
@ChrisSchinnerl are the s3d tests a blocker on this PR? I've been running a full soak test for a week now without any issues, I've done 30 snapshot recoveries, of which 3 where random snapshots, not latest and all seems to work as expected. I'm doing constant random data integrity checks without any failures so far. I'm ramping up the chaos now and trying to test more exotic scenarios but the snapshot base functionality and mechanism seem solid.

sia/persist/sqlite/init.sql
```diff
@@ -135,9 +136,26 @@ CREATE TABLE object_parts (
...
+);
+CREATE INDEX orphaned_objects_gen_idx ON orphaned_objects(orphaned_at_gen);
+
+CREATE TABLE snapshots (
+    id INTEGER PRIMARY KEY,
+    created_at INTEGER NOT NULL,
+    sia_object_id BLOB,
+    gen INTEGER NOT NULL,
+    gen_completed INTEGER,
+    object_count INTEGER NOT NULL DEFAULT 0
```
ChrisSchinnerl: No need to default here. We pass it every time anyway.

sia/persist/sqlite/snapshots.go
```diff
@@ -0,0 +1,177 @@
...
+	"go.sia.tech/core/types"
+)
+
+// CheckSnapshotObject reports whether a snapshot references the given Sia
+// object and whether any snapshot upload is still in flight. Both are
+// evaluated in a single transaction so a snapshot recorded concurrently is
+// always reported as known or in flight.
+func (s *Store) CheckSnapshotObject(objectID types.Hash256) (known, inFlight bool, err error) {
+	err = s.transaction(func(tx *txn) error {
+		if err := tx.QueryRow("SELECT EXISTS(SELECT 1 FROM snapshots WHERE sia_object_id = $1)", sqlHash256(objectID)).Scan(&known); err != nil {
```
ChrisSchinnerl: Should be able to merge those queries into a single lookup

```go
func (s *Store) CheckSnapshotObject(objectID types.Hash256) (known, inFlight bool, err error) {
	err = s.transaction(func(tx *txn) error {
		return tx.QueryRow(`
			SELECT EXISTS(SELECT 1 FROM snapshots WHERE sia_object_id = $1),
			       EXISTS(SELECT 1 FROM snapshots WHERE sia_object_id IS NULL)`,
			sqlHash256(objectID)).Scan(&known, &inFlight)
	})
	return
}
```

sia/sia.go
```diff
@@ -308,53 +350,168 @@ func (s *Sia) Close() error {
...
-	if destPath == "" {
-		return errors.New("empty destination path")
-	} else if !filepath.IsAbs(destPath) {
-		return fmt.Errorf("destination path must be absolute: %q", destPath)
+// CreateSnapshot records a snapshot, backs up the database to a temporary
+// file, uploads it gzip compressed to Sia as a tagged snapshot object, pins
+// it, and records the object ID on the snapshot. The temporary file is removed
+// once the upload completes. On failure the snapshot and any pinned object are
+// rolled back.
+func (s *Sia) CreateSnapshot(ctx context.Context) (_ s3.Snapshot, err error) {
```
ChrisSchinnerl, line 398: Probably want to make sure we remove the 1 connection limit on our conn pool to make sure this doesn't block uploads and downloads.

sia/sia.go
```diff
@@ -308,53 +350,168 @@ func (s *Sia) Close() error {
...
-		if err := s.sdk.PruneSlabs(ctx, api.WithBefore(time.Now().Add(-time.Hour))); err != nil {
-			s.logger.Error("failed to prune slabs after processing orphans", zap.Error(err))
-		} else {
-			s.logger.Info("finished pruning orphaned slabs from Sia network", zap.Duration("elapsed", time.Since(start)))
+		if prune {
+			s.logger.Info("pruning orphaned slabs")
+			start := time.Now()
+			// slabs are pinned before their object, so anything newer than
+			// pinDeadline may still belong to an upload that is retrying its pin
+			if err := s.sdk.PruneSlabs(ctx, api.WithBefore(time.Now().Add(-pinDeadline))); err != nil {
```
ChrisSchinnerl: We don't need to call this after processing orphans anymore now that deleting an objects implicitly deletes associated slabs if unused. So I'd probably put this on a slow loop that calls it once a day to clean up potential edge cases.

sia/sia.go
```diff
@@ -472,6 +629,14 @@ func (s *Sia) syncMetadataLoop(ctx context.Context) {
...
+// processing again. A sync that did not reach the end of the event stream
+// cannot prove that every snapshot on the network is known locally.
+func (s *Sia) syncFailed(msg string, fields ...zap.Field) {
+	s.logger.Error(msg, fields...)
+	s.synced.Store(false)
+}
+
 // syncMetadata fetches object events from the indexer since the last sync
 // and applies metadata updates to local objects.
 func (s *Sia) syncMetadata(ctx context.Context) { //nolint:revive
```
ChrisSchinnerl, line 826: We should probably run this a lot more often now. e.g. every minute since we rely a lot more on it now.

ChrisSchinnerl (changes requested):

sia/sia.go
peterjan: Ok, I was putting this off for the time being because there's a couple of approaches and each comes with its own set of tradeoffs. I'm testing what I think is the best solution on my soak node, results look very promising but I'll give it some more time and I'll ramp up the concurrent writers to see the effects on WAL growth, which was is the only good reason not to use VACUUM INTO imo, looks like it's not an issue at all but I'm waiting on more data.

peterjan:
@ChrisSchinnerl I have been testing this on my node for a week now. It's a 200GB node with roughly 60k files to which I'm constantly writing and reading and performing integrity checks on its data. It has now successfully recovered from a snapshot 30 times, of which 5 were stale restores. I also randomly crash the node to see if it cleanly recovers, it survived 45 crashes by now without a single file mismatch. 

I've redeployed the latest changes and I'm now mostly focusing on the backup mechanism with increased write concurrency to verify snapshots work well under continuous load. Everything is looking good though, I think we are now at a point where we can merge and mark it experimental.

ChrisSchinnerl (dismissed):

sia/persist/sqlite/snapshots.go
```diff
@@ -0,0 +1,177 @@
...
+// MarkSnapshotPinned once the backup is uploaded.
+func (s *Store) CreateSnapshot() (snap s3.Snapshot, gen int64, err error) {
+	err = s.transaction(func(tx *txn) error {
+		if err := tx.QueryRow("UPDATE global_settings SET snapshot_gen = snapshot_gen + 1 RETURNING snapshot_gen").Scan(&gen); err != nil {
+			return err
+		}
+		return tx.QueryRow(`
+			INSERT INTO snapshots (created_at, gen, object_count)
+			VALUES ($1, $2, (SELECT stat_value FROM stats WHERE stat = $3))
+			RETURNING id, created_at, object_count`, sqlTime(time.Now()), gen, statUploadedObjects).Scan(&snap.ID, (*sqlTime)(&snap.CreatedAt), &snap.ObjectCount)
```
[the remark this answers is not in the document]
peterjan: not worth doing imo, it's purely informational

sia/sia.go
```diff
@@ -376,6 +376,8 @@ func (s *Sia) CreateSnapshot(ctx context.Context) (_ s3.Snapshot, err error) {
 			// context, otherwise the object stays pinned with no local record
 			if dErr := s.sdk.DeleteObject(context.WithoutCancel(ctx), pinned); dErr != nil && !isObjectNotFound(dErr) {
 				s.logger.Error("failed to unpin snapshot object during rollback", zap.Stringer("objectID", &pinned), zap.Error(dErr))
+				// block orphan processing until we're synced
+				s.synced.Store(false)
```
ChrisSchinnerl: Do we even need this whole branch here? I think we can

1. delete the `if pinned != (...) {` with the `s.sdk.DeleteObject` call
2. drop the call to `MarkSnapshotPinned` in line 448 altogether
3. instead trigger the sync loop
4. have the sync loop finish marking the snapshot pinned

The only thing that can fail after pinning is `MarkSnapshotPinned`. If we remove it then a successful pin equals a successful backup with nothing to unpin anymore. The sync loop will handle the rest.

ChrisSchinnerl (changes requested):

sia/sia.go
peterjan: Is that better? If we drop `MarkSnapshotPinned` we leave a snapshot row with a `NULL` `sia_object_id`. That's quite annoying for a couple of reasons because in the sync loop we now have to match the event to the inflight snapshot, I guess we could use generation stamp for that but still. It also opens a new race because the sync loop might see the upload before we pin the object, so between upload and pin. Another side effect is that listing snapshots becomes eventually consistent because it filters out NULL sia object IDs. 

I'll look into it and try and write it out but I fear this is moving the complexity elsewhere for little to no benefit.

edit: it might be better, still working through the diff to make sure I haven't missed anything, it's a little sad to change this now after a full week of successful testing so want to make sure

ChrisSchinnerl: We could add a unique random identifier to the metadata when we create the snapshot. The loop shouldn't see an object that hasn't been pinned yet.

sia/sia.go
```diff
@@ -435,20 +433,19 @@ func (s *Sia) CreateSnapshot(ctx context.Context) (_ s3.Snapshot, err error) {
 	if err := s.sdk.Upload(ctx, &obj, pr); err != nil {
 		return s3.Snapshot{}, fmt.Errorf("failed to upload snapshot: %w", err)
 	}
-	pinned = obj.ID()
 
 	if err := s.sdk.PinObject(ctx, obj); err != nil {
+		// the pin may have landed on the indexer even though the request
+		// failed, block orphan processing until the sync loop reconciles
+		s.synced.Store(false)
```
ChrisSchinnerl: We still have a race here. You defer deleted the incomplete snapshot and the pin might have gone through. So that snapshot gets adopted again. Similar thing for when we eventually implement deletion. You can't just delete a snapshot from the database without risking it being adopted again. 

I think we need to implement a similar system as the one we use for file contracts on the chain in indexd by adding a state to the snapshots.

e.g. state 0 means it was created but not uploaded (set on creation)
state 1 means it's awaiting to be pinned (set before calling Pin)
state 2 means we observed it being pinned (set by sync loop)
state 3 means it is awaiting deletion (set by whoever wants to delete it)

If we roll back the snapshot in state 0 we can just delete it and on startup we can just delete all snapshots in state 0 as well.
In state 1 and 2 we have to set it to state 3 for deletion.
In state 3 we need to unpin the snapshot periodically and either delete it from the database when the indexer reports the object doesn't exist or when the sync loop reports deletion

ChrisSchinnerl (changes requested):

sia/sia.go
peterjan: Ok, I implemented it pretty much as you proposed. A snapshot now moves through created, pinning, pinned and deleting. The row flips to pinning right before the pin request goes out, so from the moment a pin can possibly reach the indexer there's always a row claiming the object. Had to add two more columns though to catch a race on startup and also another involving deletions.

peterjan: @ChrisSchinnerl ran into a little snag involving the object event stream. I think we have a bug there that causes the client to miss out on an event, which is a problem for all clients, not just us. Considering we now don't synchronously mark the object pinned but rely on the sync this opens up a race.

The second cutoff we added recently actually narrows the problem but only when the transaction commits in under a second. I wrote a test that reproduces it on master. It blocks a pin on a slab lock so its transaction spans a second boundary, hands a cursor to a client in the meantime, then asserts the committed event still gets delivered and it fails.

Another "issue" we have currently is that we treat an empty events list as being caught up, which is not technically true anymore because of the second cutoff, opening up all kind of race windows. That's on us though for relying on that. **I think the real fix is a commit ordered cursor rather than a time ordered one**, either a sequence assigned at commit or stamping at commit time. Want me to open an issue with the test attached? This is a bug that you could easily reproduce on the mobile client providing you have two devices, it would look like a refresh bug.

peterjan:
I'm converting this to `DRAFT` until we find a solution for indexd's event cursor.

sia/sia.go
```diff
@@ -466,68 +807,175 @@ func (s *Sia) syncMetadataLoop(ctx context.Context) {
 		select {
 		case <-ctx.Done():
 			return
+		case <-s.syncWake:
+			s.syncMetadata(ctx)
 		case <-t.C:
 			s.syncMetadata(ctx)
 		}
 	}
 }
```
chris124567: nit:
```
-       // sync once on startup
-       s.syncMetadata(ctx)
-
        for {
+               s.syncMetadata(ctx)
+
                select {
                case <-ctx.Done():
                        return
                case <-s.syncWake:
-                       s.syncMetadata(ctx)
                case <-t.C:
-                       s.syncMetadata(ctx)
                }
        }
 }
```

sia/persist/sqlite/objects.go
```diff
@@ -819,11 +819,21 @@ func (s *Store) AllFilenames() (filenames []string, err error) {
...
+// OrphanedObjects returns up to limit object IDs eligible for unpinning. An
+// id is withheld while any snapshot may reference it: the snapshot existed
+// when the id was orphaned and the id was first referenced before the
+// snapshot completed. An in-flight snapshot has no
+// completion generation yet and withholds every id orphaned since it started.
 func (s *Store) OrphanedObjects(limit int) (ids []types.Hash256, err error) {
 	err = s.transaction(func(tx *txn) error {
 		ids = ids[:0] // reuse same slice if transaction retries
-		rows, err := tx.Query("SELECT sia_object_id FROM orphaned_objects LIMIT $1", limit)
+		rows, err := tx.Query(`
```
chris124567, line 863: Do we need an index or something for this / do we have benchmarks

sia/sia.go
```diff
@@ -308,16 +388,211 @@ func (s *Sia) Close() error {
...
+		if rErr := os.Remove(path); rErr != nil && !errors.Is(rErr, os.ErrNotExist) {
+			s.logger.Warn("failed to remove snapshot backup file", zap.String("path", path), zap.Error(rErr))
+		}
+	}
+
+	tmp := filepath.Join(s.directory, TmpDirectory, fmt.Sprintf("snapshot-%x.tmp", frand.Bytes(8)))
+	if err := s.store.Backup(ctx, tmp); err != nil {
+		return s3.Snapshot{}, fmt.Errorf("failed to create backup: %w", err)
+	}
+	defer removeFile(tmp)
```
chris124567: nit: inline this

sia/persist/sqlite/objects.go
peterjan: Turns out we did! Added `BenchmarkOrphanedObjects` that improved the query 18x, going from well above 2s to around 100ms using an extreme benchmark scenario involving millions of orphans and hundreds of snapshot rows.

sia/persist/sqlite/init.sql
```diff
@@ -137,8 +138,32 @@ CREATE TABLE object_parts (
...
+    sia_object_id BLOB, -- backup object on the network, set before the pin is issued
+
+    gen INTEGER NOT NULL, -- generation the snapshot started at
+    gen_completed INTEGER, -- generation it completed at, NULL until pinned
+
+    state INTEGER NOT NULL, -- lifecycle state, values defined in snapshots.go
+    deleting_since INTEGER -- when the state became deleting
+);
+
+CREATE INDEX snapshots_gen_idx ON snapshots(gen, gen_completed);
```
chris124567: Is `snapshots_gen_idx` being used by sqlite in practice?  I tried `EXPLAIN ANALYZE` with and without it on different amounts of data and nothing changed

sia/sia.go
```diff
@@ -311,16 +391,209 @@ func (s *Sia) Close() error {
...
+		return false
+	}
+
+	var marked bool
+	for _, snap := range snapshots {
+		if ctx.Err() != nil {
+			return false
+		}
+		id := snap.ObjectID
+		_, err := s.sdk.Object(ctx, id)
```
chris124567, line 531: If the pinning was inflight wouldn't we want to do something about it before we look up the object with the SDK rather than after?

sia/persist/sqlite/init.sql
peterjan: Thanks for catching that, it was a leftover from `7b7d5c3`

sia/sia.go
peterjan: Not sure I fully understand what you are proposing here, but that flag is only
a veto on the rollback, not a reason to skip the lookup. The sync loop only picks up the pinned object after a full second while this object lookup will see it immediately and complete the snapshot. 

edit: I think this is moot now so I'll resolve it

sia/persist/sqlite/init.sql
```diff
@@ -36,7 +36,8 @@ CREATE TABLE sia_objects (
     encrypted_metadata BLOB,
     metadata_signature BLOB NOT NULL,
     created_at INTEGER NOT NULL,
-    updated_at INTEGER NOT NULL
+    updated_at INTEGER NOT NULL,
+    created_at_gen INTEGER NOT NULL DEFAULT 0 -- snapshot generation the id was first referenced at, kept on upsert
```
ChrisSchinnerl, line 40: Why was this reverted? There is a resolved comment about `created_at_gen`, `orphaned_at_gen` and `created_at_gen` to be important enough to be explicit about setting it, then you removed the default and now it is back.

sia/sia.go
```diff
@@ -22,13 +26,26 @@ import (
...
+	// from the Sia network.
+	pruneSlabsInterval = 24 * time.Hour
+
+	// syncMetadataInterval is the interval at which object metadata is synced
+	// from the indexer.
+	syncMetadataInterval = time.Minute
+
+	// snapshotObserveTimeout is how long CreateSnapshot waits for the sync
+	// loop to observe the pin before returning the snapshot unlisted.
+	snapshotObserveTimeout = 3 * time.Second
```
ChrisSchinnerl, line 47: Why block at all? `CreateSnapshot` successfully called `PinObject` at that point. No need to block anymore.

sia/sia.go
```diff
@@ -143,6 +164,26 @@ type Sia struct {
...
+	orphanWake chan struct{}
+	syncWake   chan struct{}
+
+	// snapshotConfirmDelay is how long after a snapshot is marked for deletion
+	// a not found reply from the indexer is distrusted, in nanoseconds
+	snapshotConfirmDelay atomic.Int64
+
+	// snapshotObserveTimeout is how long CreateSnapshot waits for the sync
+	// loop to observe the pin, in nanoseconds
+	snapshotObserveTimeout atomic.Int64
```
ChrisSchinnerl, line 179: Why atomics? Those should only be set once and then never updated?

sia/sia.go
```diff
@@ -311,16 +391,209 @@ func (s *Sia) Close() error {
...
+	if err := s.sdk.PinObject(ctx, obj); err != nil {
+		return s3.Snapshot{}, fmt.Errorf("failed to pin snapshot: %w", err)
+	}
+	snap.SiaObjectID = obj.ID()
+
+	// the sync loop completes the record once the pin's event is published,
+	// wait for it so the snapshot is listed when the call returns
+	s.wakeSyncLoop()
+	deadline := time.Now().Add(time.Duration(s.snapshotObserveTimeout.Load()))
+	for time.Now().Before(deadline) {
```
ChrisSchinnerl, line 496: This is what I meant in the other comment. If you want to make this more deterministic update the state to "pinned" here and drop the loop + the `wakeSyncLoop` calls.
If a delay is fine drop the loop and keep a single `wakeSyncLoop`.

But waiting for up to 3 seconds here seems like the worst of both worlds.

sia/sia.go
```diff
@@ -311,16 +391,209 @@ func (s *Sia) Close() error {
...
+	}
+	return snap, nil
+}
+
+// reconcileSnapshots resolves snapshots awaiting their pin against the
+// indexer. An object the indexer holds completes its record. One it does not
+// hold is marked for deletion, unless this process is still pinning it, so the
+// deletion pass can confirm the object is really gone before the record stops
+// withholding orphans. It reports whether every snapshot was resolved.
+func (s *Sia) reconcileSnapshots(ctx context.Context) bool {
```
ChrisSchinnerl, line 518: Do we need this? From what I can tell we got a race here between `PinningSnapshots` and `sdk.Object` which was patched by a slightly hacky `s.pinning` map. 

But I think we could avoid all of this by changing the `deleting_since` into `state_since` to indicate how long a snapshot has been in a given state for. Then update `SnapshotsForDeletion` to return all snapshots that have either been stuck in pinning or deletion for more than an hour.

None of the deletions really need to happen when we finish syncing. In the worst case we keep orphans around for a bit too long but that's fine.

sia/objects/objects.go
```diff
@@ -110,3 +132,52 @@ type OrphanedFile struct {
...
+// PinningSnapshot identifies a snapshot awaiting confirmation that its backup
+// object reached the indexer.
+type PinningSnapshot struct {
+	ID       int64
+	ObjectID types.Hash256
+}
+
+// DeletingSnapshot identifies a snapshot marked for deletion and when it was
+// marked.
+type DeletingSnapshot struct {
```
ChrisSchinnerl, line 145: If the only point of this type is to return `Since` for checking, why not filter in the query and just return `types.Hash256`? Saves you the type

ChrisSchinnerl (changes requested):

sia/persist/sqlite/init.sql
peterjan: Hm it seems like it snuck in with `6ffe1de`, sorry about that. All of your comments trace back to that one commit by the way. I must've ran a `simplify` on it without paying enough attention.

sia/sia.go
peterjan: I removed it but I really hate it that the API can show an empty snapshots list immediately after creation.

sia/sia.go
peterjan: A test was updating it mid-test. I got rid of it though.

sia/sia.go
peterjan: Hm your `state_since` suggestion is really nice, that collapses everything and is just overall a much more useful column to have. `reconcileSnapshots` was a third attempt at fixing the problem where you want to know whether an in-flight pin landed. I had a mutex in the syncer first, then a `stale` column on snapshots  and then finally decided to ask the indexer instead. I updated it to your suggestion. I think we have to gate deletions on being synced though, I added that.

ChrisSchinnerl (dismissed):

chris124567:
@peterjan Courtesy of Claude:

```
diff --git a/sia/sia.go b/sia/sia.go
index 51335fa..1117f33 100644
--- a/sia/sia.go
+++ b/sia/sia.go
@@ -471,7 +471,9 @@ func (s *Sia) CreateSnapshot(ctx context.Context) (_ s3.Snapshot, err error) {
    if err := s.sdk.PinObject(ctx, obj); err != nil {
        return s3.Snapshot{}, fmt.Errorf("failed to pin snapshot: %w", err)
    }
-   if err := s.store.MarkSnapshotPinned(obj.ID()); err != nil {
+   // the sync loop completes the record itself if it consumes the pin's
+   // event first, in which case the row is no longer awaiting its pin
+   if err := s.store.MarkSnapshotPinned(obj.ID()); err != nil && !errors.Is(err, objects.ErrSnapshotNotFound) {
        return s3.Snapshot{}, fmt.Errorf("failed to mark snapshot pinned: %w", err)
    }
    snap.SiaObjectID = obj.ID()
diff --git a/sia/snapshots_test.go b/sia/snapshots_test.go
index 0cbcfb9..2aeff2b 100644
--- a/sia/snapshots_test.go
+++ b/sia/snapshots_test.go
@@ -3,6 +3,7 @@ package sia_test
 import (
    "bytes"
    "compress/gzip"
+   "context"
    "encoding/json"
    "errors"
    "io"
@@ -177,6 +178,68 @@ func TestCreateSnapshot(t *testing.T) {
    }
 }
 
+// pinHookSDK runs afterPin once the wrapped SDK has pinned an object, before
+// PinObject returns to the caller.
+type pinHookSDK struct {
+   *testutil.MemorySDK
+   afterPin func(obj sdk.Object)
+}
+
+func (s *pinHookSDK) PinObject(ctx context.Context, obj sdk.Object) error {
+   if err := s.MemorySDK.PinObject(ctx, obj); err != nil {
+       return err
+   }
+   s.afterPin(obj)
+   return nil
+}
+
+// TestCreateSnapshotCompletedBySync verifies that a snapshot whose pin event
+// is consumed by a metadata sync before CreateSnapshot marks it pinned is
+// still reported as created and is not rolled back. The sync loop runs
+// independently of the request and completes the record itself when it sees
+// the event first.
+func TestCreateSnapshotCompletedBySync(t *testing.T) {
+   memSDK := testutil.NewMemorySDK()
+   hooked := &pinHookSDK{MemorySDK: memSDK}
+   backend, store := testutil.NewBackend(t, testutil.WithSDK(hooked))
+
+   // the sync loop consumes the pin's event while the request is still
+   // inside PinObject, completing the record before the request can
+   var synced bool
+   hooked.afterPin = func(obj sdk.Object) {
+       memSDK.SetEvents([]sdk.ObjectEvent{snapshotEvent(t, memSDK, obj.ID(), time.Now().Truncate(time.Second))})
+       backend.SyncMetadata(t.Context())
+       synced = true
+   }
+
+   snap, err := backend.CreateSnapshot(t.Context())
+   if err != nil {
+       t.Fatal(err)
+   } else if !synced {
+       t.Fatal("expected the sync to run during the pin")
+   }
+
+   // the snapshot is listed as pinned, not marked for deletion, and its
+   // object is still on the network
+   if snapshots, err := store.ListSnapshots(); err != nil {
+       t.Fatal(err)
+   } else if len(snapshots) != 1 {
+       t.Fatal("unexpected", len(snapshots))
+   } else if snapshots[0].ID != snap.ID {
+       t.Fatal("mismatch", snapshots[0].ID)
+   } else if snapshots[0].SiaObjectID != snap.SiaObjectID {
+       t.Fatal("mismatch", snapshots[0].SiaObjectID)
+   }
+   if ids, err := store.SnapshotsForDeletion(time.Now().Add(sia.SnapshotConfirmDelay)); err != nil {
+       t.Fatal(err)
+   } else if len(ids) != 0 {
+       t.Fatal("snapshot completed by the sync was rolled back", ids)
+   }
+   if !memSDK.Pinned(snap.SiaObjectID) {
+       t.Fatal("expected snapshot object to stay pinned")
+   }
+}
+
 // TestStuckPinningSnapshot verifies that a snapshot left awaiting its pin by a
 // dead process keeps withholding its orphans until the deletion pass confirms
 // the indexer does not hold its object, since that pin may still have been
````

File: sia/sia.go [hunk 38]
```diff
@@ -329,35 +483,97 @@ func (s *Sia) processOrphansLoop(ctx context.Context) {
...
+		case <-t.C:
+		}
+
 		s.logger.Info("pruning orphaned slabs")
 		start := time.Now()
-		if err := s.sdk.PruneSlabs(ctx, api.WithBefore(time.Now().Add(-time.Hour))); err != nil {
-			s.logger.Error("failed to prune slabs after processing orphans", zap.Error(err))
+		// slabs are pinned before their object, so anything newer than
+		// pinDeadline may still belong to an upload that is retrying its pin
+		if err := s.sdk.PruneSlabs(ctx, api.WithBefore(time.Now().Add(-pinDeadline))); err != nil {
```
chris124567, line 531: Should this run at startup given that `pruneSlabsInterval` defaults to 24h?

peterjan:
> Copilot was unable to review this pull request because the user who requested the review has reached their quota limit.

Yeah I'm aware of this one, I have one more commit locally but I'm still soak testing it. It includes the fix you mentioned.

ChrisSchinnerl (changes requested):
got a conflict

sia/sia.go
```diff
@@ -366,14 +542,55 @@ func (s *Sia) processOrphansLoop(ctx context.Context) {
...
+// and of snapshots a crash left awaiting their pin, and removes each record
+// once the indexer reports its object gone. Only records past
+// snapshotConfirmDelay are handled, so a not found reply is final. Like orphan
+// processing it waits for a metadata sync first, the sync completes a record
+// whose pin landed rather than leaving it to be reaped.
+func (s *Sia) processSnapshotDeletions(ctx context.Context, now time.Time) {
+	if !s.synced.Load() {
+		s.logger.Debug("deferring snapshot deletions until object metadata is synced")
+		return
+	}
```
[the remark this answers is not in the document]
peterjan: not really an issue imo, sync loop runs every minute while we only delete things that have been pinning for an hour

ChrisSchinnerl (approved):
got no more conflict

peterjan:
> got no more conflict

I'm sure an astra run finds 5 more race conditions that 'll never play out.

chris124567:
edit: nvm

chris124567 (approved):

ChrisSchinnerl (approved):

PR #216, opened 2026-06-24
Log errors caused by clients as DEBUG instead of ERROR to avoid spam

ChrisSchinnerl:
While testing video streaming via external storage on Nextcloud I noticed some unnecessary spam in the logs.

One source of spam is all errors in handlers being logged on the error level even if they are not 5xx codes. e.g. bad requests or in Nextcloud's case 404s resulting from Nextcloud attempting to figure out if an object is a folder or file.

The other source is the client dropping range requests. Since there is no easy way to check for a broken connection I settled on assuming that no error that happens when calling Write on the http response body is worth to be logged as an error.

### DIFF
--- a/.changeset/dont_log_clients_disconnecting_as_errors.md
+++ b/.changeset/dont_log_clients_disconnecting_as_errors.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+# Don't log clients disconnecting as errors.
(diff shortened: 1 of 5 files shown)


chris124567 (approved):
`errTrackingReader` isn't pretty but I guess it's the only way to distinguish reader error from writer error unless we want to implement even more code ourselves

ChrisSchinnerl:
> `errTrackingReader` isn't pretty but I guess it's the only way to distinguish reader error from writer error unless we want to implement even more code ourselves

Yeah I changed it a bunch of times. First I wanted to check errors but there is no good way to do that across platforms since it's all syscall specific. Then I implemented `io.Copy` and realized the fast paths are the reason we can't have nice things and I didn't want to give up on those. A wrapper seemed like the only cross platform way to distinguish errors.

peterjan (approved):

PR #219, opened 2026-06-24
Add `snapshot` command

peterjan:
This PR adds a CLI command to create, list, delete, and restore snapshots.

`s3d snapshots create/list/delete` wrap the admin API. `list --remote` enumerates the snapshots stored on the Sia network, reading the app key from the local database rather than the admin API, so it works against a stopped daemon. `restore` fetches a backup from Sia and writes it to the data directory, which is what makes a lost database recoverable from the network alone.

References https://github.com/SiaFoundation/s3d/issues/202

PR #221, opened 2026-06-25
Add backup and flush commands

chris124567:
Re: #202

### DIFF
--- a/.changeset/backup_and_flush_commands.md
+++ b/.changeset/backup_and_flush_commands.md
@@ -1,0 +1,5 @@
+---
+default: minor
+---
+
+# Added `flush` and `backup` CLI commands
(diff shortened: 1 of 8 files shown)


peterjan:
we're competing for the `backup` command it seems like 😅 https://github.com/SiaFoundation/s3d/pull/219

chris124567:
Oops... will change this to have flush only

peterjan:
> Oops... will change this to have flush only

I should have created a sub-issue and assign my self in the issue. I opened a `DRAFT` to "claim it" but that was probably not the best way to go. It was interesting though because we took totally different approaches. I think we'll end up with a combination of both. I'll update my PR to add some of the nice stuff in your PR.

PR #229, opened 2026-07-16
Normalize Sia Object

ChrisSchinnerl:
This PR updates the go sdk dependency to latest master which introduces the broken encoding for slabs.
So we also need to migrate the stored objects and while we are at it we also normalize the type so that we can more easily deal with encoding changes in the future.


chris124567 (approved):

sia/persist/sqlite/migrations_compat.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+// This file contains frozen copies of the sealed object types from
+// go.sia.tech/indexd/slabs as they were encoded before slab versioning was
+// introduced (indexd v0.3.2). Migrations use them to decode sia objects
+// stored with the legacy encoding. They must not be updated when the SDK
+// types change; a new encoding change requires a new set of legacy types and
+// a new migration.
+type (
+	legacyPinnedSector struct {
```
peterjan, line 17: `nit` I prefer our old `compatPinnedSectorv031` format

peterjan (approved):
one small thing I noticed when trying this migration out locally is that if you trigger it through a subcommand, which I did, it doesn't print anything because we pass no-op loggers - on large databases that might actually be a little annoying and might even cause users to hit ctrl+c mid migration because they think it's hanging

PR #252, opened 2026-09-03
Add bucket policies for anonymous access

ChrisSchinnerl:
Tried to move our releases over from renterd to s3d and noticed we don't have public buckets yet. This PR adds that via bucket policies. 

It won't allow arbitrary policies and instead only ones for `GetObject`, `ListBucket` and their versioned counterparts where the target is everyone (the `*` principal)

### DIFF
--- a/s3/messages.go
+++ b/s3/messages.go
@@ -509,6 +509,16 @@ type (
 
 // Types related to bucket versioning routes
 type (
+	// PolicyStatus is the response to GetBucketPolicyStatus.
+	//
+	// https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketPolicyStatus.html
+	PolicyStatus struct {
+		XMLName xml.Name `xml:"PolicyStatus"`
+		Xmlns   string   `xml:"xmlns,attr,omitempty"`
+		// IsPublic is true when the policy grants anonymous callers access.
+		IsPublic bool `xml:"IsPublic"`
+	}
+
 	// VersioningConfiguration is the S3 bucket versioning configuration
 	// document used by PutBucketVersioning and GetBucketVersioning.
 	//
--- a/s3/policy.go
+++ b/s3/policy.go
@@ -1,0 +1,350 @@
+package s3
+
+import (
+	"encoding/json/jsontext"
+	"encoding/json/v2"
+	"errors"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/SiaFoundation/s3d/s3/s3errs"
+	"go.uber.org/zap"
+)
+
+const (
+	// the only policy language version accepted; "2008-10-17" defaults differ
+	policyVersion = "2012-10-17"
+
+	policyEffectAllow = "Allow"
+	policyEffectDeny  = "Deny"
+
+	// the only principal accepted, which makes a grant available to
+	// unauthenticated callers
+	policyWildcard = "*"
+
+	policyARNPrefix = "arn:aws:s3:::"
+
+	// matches the limit AWS enforces on bucket policies
+	maxPolicySize = 20 << 10
+)
+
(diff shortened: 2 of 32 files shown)


tox/tox_test.go
```diff
@@ -156,7 +156,7 @@ func TestS3(t *testing.T) {
 	t.Run("test_s3", func(t *testing.T) {
 		runTox(t, confPath, testsDir,
 			"s3tests/functional/test_s3.py",
-			"-m", "not s3d_not_implemented and not s3d_not_supported and not bucket_logging and not encryption and not sse_s3 and not bucket_encryption and not lifecycle_transition and not tagging and not bucket_policy and not object_ownership and not checksum and not cloud_transition and not cloud_restore and not iam_user and not iam_account",
+			"-m", "not s3d_not_implemented and not s3d_not_supported and not bucket_logging and not encryption and not sse_s3 and not bucket_encryption and not lifecycle_transition and not tagging and not object_ownership and not checksum and not cloud_transition and not cloud_restore and not iam_user and not iam_account",
```
chris124567, line 159: Should be able to enable these in s3-tests now:

```
- test_bucket_policy
- test_bucketv2_policy
- test_get_bucket_policy_status
```

s3/policy.go
```diff
@@ -0,0 +1,350 @@
...
+					return BucketPolicy{}, s3errs.ErrMalformedPolicy
+				}
+				// valid, but scopes the grant to part of the bucket; s3d records
+				// grants per bucket
+				return BucketPolicy{}, s3errs.ErrNotImplemented
+			}
+		}
+
+		for _, action := range stmt.Action {
+			a, ok := supportedPolicyActions[action]
```
chris124567: nit: These are supposed to be case insensitive

> The prefix and the action name are case insensitive. For example, iam:ListAccessKeys is the same as IAM:listaccesskeys. The following examples show Action elements for different services.

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_action.html

sia/policy_test.go
```diff
@@ -0,0 +1,664 @@
...
+					return err
+				},
+			},
+			{
+				name: "a missing key is still reported as missing",
+				do: func() error {
+					_, err := c.client.GetObjectVersion(t.Context(), bucket, "missing", nil)
+					return err
+				},
+				err: &s3errs.ErrNoSuchKey,
```
chris124567: This should be `s3errs.ErrAccessDenied` if they do not have `s3:ListBucket`

s3/policy.go
```diff
@@ -0,0 +1,350 @@
...
+			// each of these restricts or inverts the grant
+			return BucketPolicy{}, s3errs.ErrNotImplemented
+		case stmt.Effect == policyEffectDeny:
+			return BucketPolicy{}, s3errs.ErrNotImplemented
+		case stmt.Effect != policyEffectAllow:
+			return BucketPolicy{}, s3errs.ErrMalformedPolicy
+		case !bool(stmt.Principal):
+			// a bucket has one owner and s3d has no other accounts or roles, so
+			// "everyone" is the only principal that can mean anything
+			return BucketPolicy{}, s3errs.ErrNotImplemented
```
chris124567, line 232: Shouldn't this be `MalformedPolicy` too?

https://stackoverflow.com/questions/60936666/getting-missing-required-field-principal-when-adding-policy-to-s3-bucket

s3/messages.go
```diff
@@ -509,6 +509,16 @@ type (
 
 // Types related to bucket versioning routes
```
chris124567, line 510: nit: either change this comment or move `PolicyStatus` out of here

chris124567 (approved):

peterjan (approved):
LGTM 🚀

PR #259, opened 2026-09-09
Store only object metadata headers

### LINKED ISSUE #257: Report server side encryption headers
`encryption` is rejected in `unsupportedBucketSubresources`, so the bucket encryption endpoints are unavailable and object responses carry no `x-amz-server-side-encryption` header. Clients and security reviews read a missing header as no encryption at rest, and enterprise procurement treats it as a yes or no question regardless of what the storage actually does. AWS, B2, MinIO and Ceph all report it.

peterjan:
Currently `metadataHeaders` keep every `X-Amz-` request header as object metadata and s3d serves it all back on GET and HEAD. This PR adds an allowlist to prevent that considering it leaks secrets on public buckets. It also eats into the metadata size budget and returns headers that might claim things we don't even support.

Master, unauthenticated GET on a public bucket:

```
HTTP/1.1 200 OK
X-Amz-Content-Sha256: ee44b2a3ae5e84e9fde84e85ddad5b90f11e43b2ee22a8e30187254b2d100ef5
X-Amz-Date: 20260916T140543Z
X-Amz-Security-Token: FwoGZXIvYXdzEBYaDFAKESTSSESSIONTOKENabc123
X-Amz-Server-Side-Encryption-Customer-Algorithm: AES256
X-Amz-Server-Side-Encryption-Customer-Key: pO3upElrwuEXSoFwCfnZPdSsmt/xWeFa0N9KgDijwVs=
X-Amz-Server-Side-Encryption-Customer-Key-Md5: DWygnHRtgiJ77HCm+1rvHw==
```

Kopia, a tool that I was testing, sends exotic headers that get accepted and then repeated on GET/HEAD:
```
PutObject with x-amz-object-lock-mode GOVERNANCE   accepted
HeadObject reports ObjectLockMode                  "GOVERNANCE"
GetObjectRetention on the same object              501
DeleteObject inside the retain window              accepted, object gone
```

### DIFF
--- a/.github/workflows/tox-test.yml
+++ b/.github/workflows/tox-test.yml
@@ -43,7 +43,7 @@ jobs:
           python -m pip install "tox>=4"
 
       - name: Clone test repo
-        run: git clone https://github.com/SiaFoundation/s3-tests.git
+        run: git clone --branch pj/sse-markers https://github.com/SiaFoundation/s3-tests.git
 
       # warm up the tox env so the Go test's `tox --skip-pkg-install` is fast
       - name: Setup test suite (create envs only)
--- a/s3/objects.go
+++ b/s3/objects.go
@@ -1313,7 +1313,7 @@ func writeGetOrHeadObjectHeaders(obj *Object, w http.ResponseWriter, r *http.Req
 		}
 
 		// older versions stored request headers as metadata, including
-		// customer encryption keys, so serve only the object's own
+		// customer encryption keys, so serve only the object's own metadata
 		if !isObjectMetadataHeader(mk) {
 			continue
 		}
--- a/s3/objects.go
+++ b/s3/objects.go
@@ -199,7 +199,7 @@ func (s *s3) routeObject(w http.ResponseWriter, r *http.Request, accessKeyID *st
 }
 
 func (s *s3) copyObject(w http.ResponseWriter, r *http.Request, accessKeyID, dstBucket, dstObject string, meta map[string]string) error {
-	source := meta["X-Amz-Copy-Source"]
+	source := r.Header.Get("X-Amz-Copy-Source")
 	log := s.logger.With(zap.String("dstBucket", dstBucket),
 		zap.String("dstObject", dstObject),
 		zap.String("source", source),
@@ -240,6 +240,7 @@ func (s *s3) copyObject(w http.ResponseWriter, r *http.Request, accessKeyID, dst
 		w.Header().Set("x-amz-copy-source-version-id", result.SourceVersionID)
 	}
 
+	setSSEResponseHeader(w)
 	etag := FormatETag(result.ContentMD5[:], int(result.PartsCount))
 	w.Header().Set("ETag", etag)
 	return writeXMLResponse(w, http.StatusOK, ObjectCopyResult{
@@ -328,6 +329,10 @@ func (s *s3) serveObject(w http.ResponseWriter, r *http.Request, accessKeyID *st
 		log.Debug("head object")
 	} else {
 		log.Debug("get object")
+	}
+
+	if err := validateSSEReadHeaders(r.Header); err != nil {
+		return err
 	}
 
 	partNumber, err := parsePartNumber(r.URL.Query().Get("partNumber"))
@@ -739,13 +744,18 @@ func (s *s3) putObject(w http.ResponseWriter, r *http.Request, accessKeyID strin
 		return s3errs.ErrKeyTooLongError
 	}
 
+	// validated here rather than in copyObject so a copy is covered as well
+	if err := validateSSEWriteHeaders(r.Header); err != nil {
+		return err
+	}
+
 	// extract metadata headers
 	meta, err := metadataHeaders(r.Header, MetadataSizeLimit)
 	if err != nil {
 		return err
 	}
 
-	if _, ok := meta["X-Amz-Copy-Source"]; ok {
+	if _, ok := r.Header["X-Amz-Copy-Source"]; ok {
 		return s.copyObject(w, r, accessKeyID, bucket, object, meta)
 	}
 
@@ -785,6 +795,7 @@ func (s *s3) putObject(w http.ResponseWriter, r *http.Request, accessKeyID strin
 	if res.VersionID != "" {
 		w.Header().Set("x-amz-version-id", res.VersionID)
 	}
+	setSSEResponseHeader(w)
 	w.Header().Set("ETag", FormatETag(res.ContentMD5[:], 0))
 	return nil
 }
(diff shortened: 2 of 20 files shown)


.github/workflows/tox-test.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           python -m pip install "tox>=4"
 
       - name: Clone test repo
-        run: git clone https://github.com/SiaFoundation/s3-tests.git
+        run: git clone --branch pj/sse-markers https://github.com/SiaFoundation/s3-tests.git
```
peterjan, line 46: reminder to revert `56ec27b5` before merging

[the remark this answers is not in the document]
peterjan: the commit msg literally reads "tmp: revert later"

ChrisSchinnerl (commented):
My problem with this is that
1. it effectively just lies to clients about the encryption since we don't apply AES
2. I'm not sure what use-case this enables apart from supporting something that other backend support. e.g. is there an app that doesn't work without it?
3. this is a massive PR for faking AES headers on a self-hosted tool for which you control the backend anyway

Looking at [Cloudflare's compatibility](https://developers.cloudflare.com/r2/api/s3/api/) it seems like they also don't support SSE. Which doesn't mean we can't have anything that Cloudflare doesn't support but if a player of their size doesn't feel like the demand is there maybe it's fine not to have it.

If we did add it I might honestly lean towards actually applying additional AES256 encryption simply for the sake that users who explicitly want that feature actually get what they want. 
e.g. by using a configurable secret that never gets sent to our backend including snapshots. So even if you used an app like Revelio to access the same account, you wouldn't be able to decrypt any of the objects.
Although the need for this still feels quite unlikely.

sia/persist/sqlite/encryption.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"database/sql"
+	"errors"
+
+	"github.com/SiaFoundation/s3d/s3/s3errs"
+)
+
+// PutBucketEncryptionConfiguration stores the serialized default encryption
+// configuration for a bucket, replacing any existing configuration.
+func (s *Store) PutBucketEncryptionConfiguration(accessKeyID, bucket, config string) error {
```
chris124567: Why do we need this if we just report AES256 regardless

chris124567 (commented):
I agree with Chris and if we do really we want to support this just for the sake of some s3 clients we should implement the absolute bare minimum

peterjan:
> My problem with this is that
> ...

Hm, that's all fair points. I've reduced the PR to what I think is worth salvaging, the metadata headers allowlist. That's actually how I ended up at SSE in the first place. 

We store every X-Amz- request header as object metadata and serve it back on GET/HEAD reqs, potentially leaking secrets on public buckets. It also eats into the metadata budget and makes s3d claim features it doesn't support because it's echoing whatever was sent on upload.

Master, unauthenticated GET on a public bucket:
```
HTTP/1.1 200 OK
X-Amz-Content-Sha256: ee44b2a3ae5e84e9fde84e85ddad5b90f11e43b2ee22a8e30187254b2d100ef5
X-Amz-Date: 20260916T140543Z
X-Amz-Security-Token: FwoGZXIvYXdzEBYaDFAKESTSSESSIONTOKENabc123
X-Amz-Server-Side-Encryption-Customer-Algorithm: AES256
X-Amz-Server-Side-Encryption-Customer-Key: pO3upElrwuEXSoFwCfnZPdSsmt/xWeFa0N9KgDijwVs=
X-Amz-Server-Side-Encryption-Customer-Key-Md5: DWygnHRtgiJ77HCm+1rvHw==
```

Kopia, a tool that I was testing, sends exotic headers that get accepted and then repeated on GET/HEAD:
```
    PutObject with x-amz-object-lock-mode GOVERNANCE   accepted
    HeadObject reports ObjectLockMode                  "GOVERNANCE"
    GetObjectRetention on the same object              501
    DeleteObject inside the retain window              accepted, object gone
```

s3/objects.go
```diff
@@ -846,17 +846,44 @@ func parseSource(source string) (bucket, object string, version VersionRequest,
...
+// metaPrefix instead.
+var objectMetadataHeaders = map[string]struct{}{
+	"Cache-Control":                   {},
+	"Content-Disposition":             {},
+	"Content-Encoding":                {},
+	"Content-Type":                    {},
+	"Expires":                         {},
+	"X-Amz-Checksum-Crc32":            {},
+	"X-Amz-Checksum-Crc32c":           {},
+	"X-Amz-Checksum-Crc64nvme":        {},
```
chris124567, line 866: There are more checksum headers: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html

peterjan: I only listed the ones we verify in https://github.com/SiaFoundation/s3d/pull/261 . Could consider supporting all though but I'll wait for Chris's review on this PR considering I'm not even sure we want this change to begin with.

peterjan: I added md5 and sha256. I think that leaves 4 or so that we don't cover.

s3/objects.go
```diff
@@ -846,17 +846,47 @@ func parseSource(source string) (bucket, object string, version VersionRequest,
...
+	"X-Amz-Checksum-Md5":              {},
+	"X-Amz-Checksum-Sha1":             {},
+	"X-Amz-Checksum-Sha256":           {},
+	"X-Amz-Checksum-Sha512":           {},
+	"X-Amz-Website-Redirect-Location": {},
```
ChrisSchinnerl: we don't support static website hosting and redirects so I'd comment this line out but keep it as a reminder that it theoretically should be stored but is simply not supported yet. Otherwise a PutObject could probably set this and we would replay it but don't actually support it.

Same thing for the checksums we don't support.

s3/objects.go
```diff
@@ -846,17 +846,47 @@ func parseSource(source string) (bucket, object string, version VersionRequest,
...
+}
+
+// isObjectMetadataHeader reports whether the canonically formatted header name
+// belongs with the object rather than with the request that carried it.
+func isObjectMetadataHeader(name string) bool {
```
ChrisSchinnerl, line 880: Would be nice to have a migration that prunes any metadata that shouldn't be stored from the database. Otherwise `CopyObject` will just carry those headers over.

ChrisSchinnerl (changes requested):

s3/objects.go
peterjan: Yeah I was on the fence on whether to add it or not. Figured probably ok not to but I added it 👍 to be honest I never considered `CopyObject`.

s3/objects.go
peterjan: Hm ok, I commented it out. Checksums have a PR, I'll make sure to only add those which we support.

ChrisSchinnerl (approved):

ChrisSchinnerl (changes requested):
Got conflicts now

ChrisSchinnerl (approved):

PR #264, opened 2026-09-17
Add log flags

peterjan:
This PR adds two flags, `-log.file.enabled` and `-log.stdout.enableANSI`,  to allow configuring the logger. 
The defaults stay the same. 

References https://github.com/SiaFoundation/linux/issues/129


cmd/s3d/main.go
```diff
@@ -89,6 +89,8 @@ func main() {
 	rootCmd.Usage = flagg.SimpleUsage(rootCmd, rootUsage)
 	rootCmd.StringVar(&cfg.ApiAddress, "api.s3", cfg.ApiAddress, "address to serve S3 API on")
 	rootCmd.StringVar(&cfg.ApiHttpsAddress, "api.s3.https", cfg.ApiHttpsAddress, "address to serve the S3 API on over HTTPS using a self-signed certificate (disabled if empty)")
+	rootCmd.BoolVar(&cfg.Log.File.Enabled, "log.file.enabled", cfg.Log.File.Enabled, "enable file logging")
+	rootCmd.BoolVar(&cfg.Log.StdOut.EnableANSI, "log.stdout.enableANSI", cfg.Log.StdOut.EnableANSI, "enable ANSI color codes")
```
peterjan, line 93: I was a little annoyed with this second flag. It's because `s3d config` writes the whole config struct, which writes a value for this field, so I think we need a flag to override it.

chris124567 (approved):

ChrisSchinnerl (approved):

